Issue #18, opened 2021-06-14
ExcludeColumn(columns ...string) *InsertQuery


oiime:
Hello, I was trying to exclude some columns from an insert query and noticed that method is not supported, Would it make sense to add it? excludeColumns already exists in baseQuery

vmihailenco:
Yes, it makes sense to have `ExcludeColumn` on `InsertQuery`. Please send a PR if you have a chance.

vmihailenco:
Closed by https://github.com/uptrace/bun/pull/19

PR #19, opened 2021-06-14
Add ColumnExpr and ExcludeColumn to UpdateQuery and InsertQuery

oiime:
Small change per #18 to add exclude column to InsertQuery

### DIFF
--- a/query_insert.go
+++ b/query_insert.go
@@ -83,6 +83,16 @@ func (q *InsertQuery) Column(columns ...string) *InsertQuery {
 	return q
 }
 
+func (q *InsertQuery) ColumnExpr(query string, args ...interface{}) *InsertQuery {
+	q.addColumn(schema.SafeQuery(query, args))
+	return q
+}
+
+func (q *InsertQuery) ExcludeColumn(columns ...string) *InsertQuery {
+	q.excludeColumn(columns)
+	return q
+}
+
 // Value overwrites model value for the column in INSERT and UPDATE queries.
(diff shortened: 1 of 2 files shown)


query_insert.go
```diff
@@ -83,6 +83,16 @@ func (q *InsertQuery) Column(columns ...string) *InsertQuery {
 	return q
 }
 
+func (q *InsertQuery) ColumnExpr(query string, args ...interface{}) *InsertQuery {
+	q.addColumn(schema.SafeQuery(query, args))
+	return q
+}
```
vmihailenco, line 89: Do you have an example of the query with ColumnExpr? If so we may add such query to https://github.com/uptrace/bun/blob/master/internal/dbtest/query_test.go#L396-L403 so we don't break it.

vmihailenco:
Thanks

query_insert.go
oiime: @vmihailenco I'd assume anything that requires calling a stored procedure, a static value or a built in function would benefit from ColumnExpr, eg: ColumnExpr("NOW()") or ColumnExpr("SomeStoredProcedureWithReturn()")

vmihailenco: I believe it should be `Value("column", "now()")` - see https://github.com/uptrace/bun/pull/22. I am going to remove `ColumnExpr` since it does not produce valid query.

oiime: Makes sense, can be removed I guess

PR #22, opened 2021-06-16
Remove ColumnExpr and add query tests for Value

vmihailenco:
Partially reverts https://github.com/uptrace/bun/pull/19


vmihailenco:
@oiime please let me know if there is a use case for `ColumnExpr`. I am removing it for now.

Issue #29, opened 2021-06-21
Add warning when zero time ('0001-01-01 00:00:00+00:00') is appended


vmihailenco:

rojanDinc:
Hi I'd like to contribute. To what should this warning be displayed when a zero time is appended?

vmihailenco:
Hi @rojanDinc,

We have the `warn` logger here - https://github.com/uptrace/bun/blob/master/internal/logger.go#L10. Then in schema/table.go you can add a check like:

```go
if field.Type == timeType && !field.NullZero {
    // warn user
}
```

rojanDinc:
Thanks for the pointers 👍

PR #32, opened 2021-06-24
Add warn log if field is zero valued time

rojanDinc:
Is this the only place for this check to be done? Couldn't find any other ones.

Closing #29


vmihailenco:
Thanks :+1:

PR #40, opened 2021-06-29
Add sqliteshim package

tie:
This PR adds sqliteshim package that selects SQLite driver implementation (currently either [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite) or [github.com/mattn/go-sqlite3](https://pkg.go.dev/github.com/mattn/go-sqlite3)) based on the CGO_ENABLED/GOOS/GOARCH target.


driver/sqliteshim/sqliteshim.go
```diff
@@ -0,0 +1,42 @@
...
+const (
+	// ShimName is the name of the shim database/sql driver registration.
+	ShimName = "sqliteshim"
+
+	// DriverName is the name of the database/sql driver. Note that
+	// the value depends on the build target.
+	DriverName = driverName
+
+	// HasDriver indicates that SQLite driver implementation is available.
+	HasDriver = !needsCgo || (needsCgo && usesCgo)
```
vmihailenco, line 29: I wonder why this is written in a such manner? Why not just have a bool var?

Also what do you think about changing these 3 vars to be methods? It is easier to make backwards-compatible changes this way.

vmihailenco:
Looks pretty good overall - thanks :+1:

vmihailenco:
:+1: I will update tests and docs to use the shim. Thanks!

tie:
@vmihailenco thanks! I’m already working on tests to use both shim and modernc directly, as well as some minor improvements to cleanup (e.g. asserting no errors in db.Close).

vmihailenco:
Okay, please pull latest changes so there are no conflicts :)

PR #43, opened 2021-06-30
insert: mysql does not support insert with alis

j2gg0s:
Example in my repo: https://github.com/j2gg0s/bun/blob/mysql/example/basic/main.go#L83

### DIFF
--- a/db.go
+++ b/db.go
@@ -486,6 +486,10 @@ func (db *DB) supportsDeleteTableAlias() bool {
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks - this is one way to do it. We also have [feature-flagging](https://github.com/uptrace/bun/blob/master/dialect/feature/feature.go#L9-L20) for features that are statically supported/not supported in some databases. This way we can enable the feature in the dialects that support it which should be more manageable (I think).

I will send a PR to clean this up. Merging meanwhile.

PR #51, opened 2021-07-01
Allow ptr of time.Time as SoftDelete filed

j2gg0s:
And we still does not allow delete with nil pointer model, such as :
```
-       _, err = db.NewDelete().Model(video1).Where("id = ?", video1.ID).Exec(ctx)
+       _, err = db.NewDelete().Model((*Video)(nil)).Where("id = ?", video1.ID).Exec(ctx)
```

Will we support this?


vmihailenco:
>Will we support this?

I guess we should. We probably should start adding tests for such features :)

PR #55, opened 2021-07-02
pk not int, without auto increment.

j2gg0s:


vmihailenco:
Thanks. I've removed float32 and float64 since they probably can't be used primary keys. Also cleaned up the test a bit.

Issue #41, opened 2021-06-29
Make migrations more explicit and generic


Codebreaker101:
I've started to use bun migrations and I have a few suggestions for possible improvements and would like to hear your opinions on them. And if you approve some of them I could start working on a PR. 

1. Remove `context` and `bun.DB` parameters from `migrations.CreateGo` or `migrations.CreateSQL` function since they are not used. Is it planned for the future?
2. Make migration directory configurable. Currently, it gets the caller directory, which means I have to alias `Create(.*)` functions in the directory where my migrations are located since I'm calling Create function from some other directory.
3. Make package name configurable for `CreateGo` migration (and/or item 6.)
``` go
func CreateGo(directory, name, package string) error
func CreateSQL(directory, name string) error
```

4. Replace logging with `fmt.Printf` with user configurable logger. Maybe global log function for migration package?  Something like:
``` go
var log = func(format string, a ...interface{}) {
	fmt.Printf(format, a...)
}

func SetLogFn(fn func(format string, a ...interface{})) {
	log = fn
}
```
5. Add more code to generated `go` template like the one bellow since this is a MVP for go migration file.  
``` go
package {{ package_name }}

import (
	"context"

	"github.com/uptrace/bun"
)

func init() {
	Migrations.MustRegister(func(ctx context.Context, db *bun.DB) error {
		fmt.Print(" [up migration] ")
		return nil
	}, func(ctx context.Context, db *bun.DB) error {
		fmt.Print(" [down migration] ")
		return nil
	})
}
```
6. Add option for adding user template for go and sql migration file creation.
7. Add a check before migrate if the required migration tables exist, if not, initialize. Quick and dirty way is to run Init(...) on every migration. Might be cheaper to execute then checking first. But speed is not an issue here. 
8. `Status` function should return a struct containing migration status and not print to console.

vmihailenco:
All of that sounds good :+1: We just need to figure out the details :) I think it will be easier if you split these changes into multiple PRs so we can merge it faster and there is a sense progress being made.

>function since they are not used. Is it planned for the future?

Not really, but I think they can be useful at some point (especially context). I am not sure, but since it is a public API I would not change the signature without a good reason. See below how I suggest to extend the API with options.

>Make package name configurable for CreateGo migration (and/or item 6.)

What about keeping the existing signature and instead adding an option:

```
func CreateGo(ctx context.Context, db *bun.DB, name string, opts ...GoOption) error

[...]

PR #58, opened 2021-07-05
Migration status returns struct. 

Codebreaker101:
References point 8 of the #41 issue.


migrate/migrations.go
```diff
@@ -282,50 +283,112 @@ func (m *Migrations) MarkCompleted(ctx context.Context, db *bun.DB) error {
...
 }
 
-func (m *Migrations) Status(ctx context.Context, db *bun.DB) error {
-	allMigrations := m.sortedMigrations()
-	if len(allMigrations) > 0 {
-		fmt.Printf("In total you have %d migrations: %s\n",
-			len(allMigrations), migrationNames(allMigrations))
+type Status struct {
+	UpToDate bool
+	Locked   bool
```
vmihailenco, line 288: Let's remove these 2 vars?
- UpToDate seems to be unused and `len(s.NewMigrations) == 0` is good enough
- Locked seems unneeded too. I believe `Status` should return an error when table can't be locked.

migrate/migrations.go
```diff
@@ -282,50 +283,112 @@ func (m *Migrations) MarkCompleted(ctx context.Context, db *bun.DB) error {
...
+
+	Migrations           []Migration
+	NewMigrations        []Migration
+	LastMigrationGroup   []Migration
+	LastMigrationGroupID int64
+}
+
+func (s Status) String() string {
```
vmihailenco, line 296: Interesting idea - I like it!

vmihailenco:
Looks pretty good - one small point about `Locked` before we can merge this.

migrate/migrations.go
Codebreaker101: Sure. But, I would like to at least add `var ErrLocked = errors.New("migrate: migrations table is locked")` so that one could check if the table is locked. 

> It is an internal detail, not public interface.

Lock and Unlock functions are exported and used in the [bun-starter-kit](https://github.com/go-bun/bun-starter-kit/blob/0447f252b50e04e33ffeb8285092a507e06d1f80/cmd/bun/main.go#L124-L149), so it seems to me it kind of is. Unless I'm missing something.

vmihailenco: >Lock and Unlock functions are exported and used in the bun-starter-kit, so it seems to me it kind of is. Unless I'm missing something.

I mean that `Status` function may or may not lock the migrations so it is best to not make `Locked` part of the `Status` API. For example, in future we may decide to wait until lock is released.

>But, I would like to at least add 

Let's add it :+1:

PR #61, opened 2021-07-05
add custom logger

Codebreaker101:
This PR allows setting one's own logger function.

### DIFF
--- a/migrate/migrations.go
+++ b/migrate/migrations.go
@@ -18,6 +18,14 @@ import (
 	"github.com/uptrace/bun"
 )
 
+var log = func(format string, a ...interface{}) {
+	fmt.Printf(format, a...)
+}
+
+func SetLogFn(fn func(format string, a ...interface{})) {
+	log = fn
+}
+
 type Command struct {
 	Name string
 }
@@ -125,7 +133,7 @@ func (m *Migrations) Discover(fsys fs.FS) error {
 			return nil
 		}
 
-		return errors.New("not reached")
+		return errors.New("migrate: not reached")
 	})
 }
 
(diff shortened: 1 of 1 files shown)


migrate/migrations.go
```diff
@@ -18,6 +18,14 @@ import (
...
 )
 
+var log = func(format string, a ...interface{}) {
+	fmt.Printf(format, a...)
+}
+
+func SetLogFn(fn func(format string, a ...interface{})) {
+	log = fn
+}
+
```
vmihailenco, line 28: Can you please change this to be an interface with a single method?

```go
type Printer interface {
    Printf(format string, args ...interface{})
}
```

It is easier to extend/change interface-based API in future (e.g. add/remove methods).

vmihailenco:
@Codebreaker101 gentle ping. What do you think about https://github.com/uptrace/bun/pull/61#discussion_r663923886?

Codebreaker101:
@vmihailenco my bad. I started a review without actually publishing my comment

migrate/migrations.go
vmihailenco: Yes, exactly that (except replacing `logger` with `printer` in few places). Ideally, it should be a field on some struct, but a global printer is fine too if that is not possible.

vmihailenco:
FYI I had to restore newlines, because otherwise the output is messed up

```
[bun]  09:42:49.610   INSERT                1.442ms  INSERT INTO bun_migration_locks ("table_name") VALUES ('bun_migrations') RETURNING "id"
[bun]  09:42:49.616   SELECT                   87µs  SELECT * FROM bun_migrations ORDER BY id DESC
rolling back group #1 with 2 migrations...rolling back migration 20210505110847... nothing to run[bun]  09:42:49.620   DELETE                 4.64ms  DELETE FROM bun_migrations WHERE (id = 2)
rolling back migration 20210505110026...  [down migration] done rolling back migration 20210505110026[bun]  09:42:49.625   DELETE                4.087ms  DELETE FROM bun_migrations WHERE (id = 1)
[bun]  09:42:49.627   DELETE                2.523ms  DELETE FROM bun_migration_locks WHERE ("table_name" = 'bun_migrations')
```

Issue #49, opened 2021-06-30
Should there be a way to migrate to a specific target?


smyrman:
I want to suggest add a method to `migrate.Migrations` to migrate to a specific group number / target. E.g.

```go
func(m *Migrations) MigrateTo(ctx context.Context, db *bun.DB, target int64) error
```

Or an example on how to achieve this without this method.

Use-case:
1. Deploy new API that require the latest schema version (e.g. `5`).
2. Find there is a bug in the API, and roll it back to a previous version that requires an earlier db schema (e.g. `3`).
3. Run to migrate to group `3` (automatically rollback any migrations with a higher group number, and apply any migrations with a lower or eqaul group number).

Assumes that migrations group numbers can be set from the source.

vmihailenco:
>Assumes that migrations group numbers can be set from the source.

Currently you can't set a group number from the source. Group numbers are generated automatically whenever you run migrations.

In theory it is possible to add some way to manually override group numbers (though nothing obvious comes to mind), but then you could also merge migrations together if they must be run as one. Or just delete the migrations you don't want to apply.

>Should there be a way to migrate to a specific target?

I am inclined to say no. At least until you come with a more detailed plan :) Right now group numbers is just an implementation detail and they don't have much meaning / purpose.

smyrman:
The motivation is to have a declarative description of the entire application deployment, including the database version, similar to what we currently do with Gorm (v1) and Gormigrate.

In our current set-up with Kustomize, we have a migration _job_ that are deployed alongside every update to our various micro-services. This job ensures that the database is migrated (up or down) to a _specific_ version that's specified as a kustomize variable. This should happen no matter the state of the initial database. For cases where a down migration _isn't safe_, then we would leave it empty (or in sine cases error).

Perhaps relying on group numbers isn't the best idea, or perhaps bun/migrations isn't the right choice (we could use bun without bun migrations). It's also not important for this to be part of the starter-kit, if included. If there was a RollbackAll (ignoring the group IDs), that _would_ probably help because we could find our _target_ in the migration list, split it in two, migrate the first half, and rollback the second half.

vmihailenco:
I guess I misunderstood. Assuming that there are 2 groups with 2 migrations in each group:

```
group1: migration1, migration2
group2: migration3, migration4
```

And you want to migrate to migration1, then Bun should rollback both groups and then apply migration1:

```shell
bun db migrate migration1
reverting group1: migration1, migration2
reverting group2: migration3, migration4
[...]

PR #67, opened 2021-07-08
Rename log and restore newlines

vmihailenco:
@Codebreaker101 PTAL


migrate/migrations.go
```diff
@@ -439,16 +440,16 @@ func (m *Migrations) genMigrationName(name string) (string, error) {
 }
 
 func (m *Migrations) runUp(ctx context.Context, db *bun.DB, migration *Migration) error {
-	m.log.Printf("running migration %s... ", migration.Name)
+	m.printer.Printf("running migration %s... ", migration.Name)
 	if migration.Up != nil {
 		if err := migration.Up(ctx, db); err != nil {
-			m.log.Printf("unable to run migration %s: %s", migration.Name, err)
+			m.printer.Printf("failure: %s\n", err)
```
vmihailenco, line 446: No need to repeat the migration twice, because the full log string is `running migration %s... failure: %s\n`

Codebreaker101:
TBH I'm not happy with the state of the logger, or in this case _status printer_. My initial PR where a user can se a logging function was primarily to be able to turn off any printing whilst keeping current functionality. Also, I was deterred from making large changes since #49 would probably break that. 

From what I can tell, this package does not have any logging, only status printing. It was my mistake for thinking that the `Printf`s were logging, hence the naming of vars and functions in my PR. If we keep in mind that this is a package, what can be done? 

Let's not make any assumptions about the user -- if a user wants any printing, they should enable it. No package should print anything unless told otherwise. It is ok for bun-starter-kit, since it is an app, but not from a migration package, since in most cases (my opinion) this package will be integrated into other apps and any printing is done by the app. If looking from the context of logging, new lines, tabs and such should be avoided since they don't work well with logging packages. 

We need a way for user to have access to migration/rollback status during the execution of the actions. To do that we could implement some kind of hooks:

``` go
// general idea 
type MigrationHooks interface {
	BeforeMigrateRun(group int64, migrations []Migration)
	AfterMigrateRun(group int64, migrations []Migration)

	BeforeMigrate(migration *Migration)
	AfterMigrate(migration *Migration, err error)
	
	BeforeRollbackRun(group int64, migrations []Migration)
	AfterRollbackRun(group int64, migrations []Migration)
	
	BeforeRollback(migration *Migration)
	AfterRollback(migration *Migration, err error)
}
```
That was, status printing can be implemented in the bun-starter-kit, any user is free to implement their own and migration package is clean from any printing.

vmihailenco:
Good default logging is an essential feature for migrate package. I understand that some users want to customize it, but the majority will not bother.

I believe that logging should be enabled by default because it helps new users understand what is happening. Once they get used to it, they may want to disable all logging. But first they must get used to it.

Perhaps we can keep all the logging from migrate package and move it to the starter-kit. That makes the starter-kit even more important/essential, but perhaps that is a good thing...

vmihailenco:
@Codebreaker101 please take another look

Codebreaker101:
@vmihailenco looking good! Migrator returns its status after run so no need for hooks. Great!

vmihailenco:
Thanks for the feedback. Merging then :)

PR #70, opened 2021-07-10
Add MarkDone and MarkUndone

vmihailenco:
Updates https://github.com/uptrace/bun/issues/49

This PR
- Adds MarkDone / MarkDone to mark individual migrations
- Adds `Migrations.Sorted` to get the list of all migrations
- Adds new option `WithMigrationDryRun` that can be used with `Migrate` and `Rollback` to mark migrations as done/undone without applying them
- Adds `SelectDoneMigrations` that returns the list of migrations stored in the database
- Adds `SelectLastGroup` that returns the last migration group

@smyrman @Codebreaker101 please review

### DIFF
--- a/migrate/migration.go
+++ b/migrate/migration.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"sort"
 	"strings"
 	"time"
 
@@ -26,6 +27,10 @@ type Migration struct {
 
 func (m *Migration) String() string {
 	return m.Name
+}
+
+func (m *Migration) IsApplied() bool {
+	return m.ID > 0
 }
 
 type MigrationFunc func(ctx context.Context, db *bun.DB) error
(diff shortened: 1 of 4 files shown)


migrate/migrator.go
```diff
@@ -174,39 +195,10 @@ func (m *Migrator) selectLastGroup(ctx context.Context) (*MigrationGroup, error)
...
-		migration.GroupID = group.ID
-		migration.Up = nil
-		if err := m.runUp(ctx, m.db, migration); err != nil {
-			return nil, err
-		}
-	}
-
-	return group, nil
+	log.Printf(
+		"DEPRECATED: bun: replace MarkCompleted(ctx) with " +
```
smyrman, line 199: Normally, dry-run means to print what will be done without making any changes.

That is, dry-run is useful, but probably should not do any changes to the DB?

smyrman (commented):
Looking at this from my phone, so not a proper review. Maybe rename DryRun as it usually means something else.

I don't know whats a better name. StateOnly maybe?

smyrman:
Btw, `migrate.StateOnly` contain all the information that `migrate.WithMigrationStateOnly` does, but is shorter and easier to read.

vmihailenco:
@smyrman all done - the API is much cleaner with `MigrationsWithStatus` :+1: PTAL

>I don't know whats a better name. StateOnly maybe?

It is rather obscure... I've renamed `WithDryRun` to `WithoutMigrationFunc` - I don't like it too much but at least it describes what is happening underneath.

migrate/migration.go
```diff
@@ -136,13 +141,72 @@ func (ms MigrationSlice) String() string {
...
+	for i := range ms {
+		if ms[i].IsApplied() {
+			applied = append(applied, ms[i])
+		}
+	}
+	sortDesc(applied)
+	return applied
+}
+
+// Unrone returns unapplied migrations in ascending order
```
Codebreaker101, line 157: Just a small typo :)

Codebreaker101:
> WithoutMigrationFunc

Maybe `SkipExecutingMigrations` would be a better name?

Additionally, what happens when one marks a migration as complete and then tries to rollback that same migration group, but it has `.down.sql` scripts associated with it? I couldn't find code that handles this situation. Should we let them run or ignore those down scripts whose up scripts were marked as complete?

vmihailenco:
Thanks for the review.

>Maybe SkipExecutingMigrations would be a better name?

Not sure :) But I've renamed it to WtihNopMigration  (nop = no operation) for now.

>Should we let them run or ignore those down scripts whose up scripts were marked as complete?

`Rollback` already accepts `WtihNopMigration` to skip running actual migrations during rollbacks. So I think we are fine?

smyrman:
I felt the `migrate.StateOnly` was quite clear; you migrate the state (but don't run the migrations). The package name _should_ be read as part of the type name in Go.

As for WtihNopMigration, the full name would then be `migrate.WtihNopMigration` when used by end-users. Would it not be better to skip the Migration name at is sort of duplicated from the package name? What about just migrate.Nop?

```go
m := migrate.NewMigrator(ms)
m.Migrate(ctx, migrate.Nop)
```
v.s.

```go
m := migrate.NewMigrator(ms)
m.Migrate(ctx, migrate.WtihNopMigration)
```

vmihailenco:
>I felt the migrate.StateOnly was quite clear

The problem is that state is rather vague and we don't use that term anywhere else. `WithMarkOnly` could be a better name with the same idea, but I am not sure...

>migrate.WtihNopMigration isn't to clear to me either...

The previous version of code used to do `migration.Up = nil`. Thus `WtihNopMigration` or `WtihNilMigration` which describes the inner mechanics.

>migrate.SkipExecutingMigrations is actually clear, it's just very long.

It is clear, but sounds a bit awkward. SkipExecutingMigrations -> WithoutRunningMigrations -> WithNoMigrations -> WithNopMigration. I don't see a clear winner and it is best to make some progress :)

smyrman:
Bice work🎉

PR #74, opened 2021-07-12
Bump deps version on last bun release

### LINKED ISSUE #56: Write/find some script that bump versions/dependencies
It might be good to make the plugin modules explicitly mark their dependency on the main bun module with the same version.

So for example in bun/dialect/pgdialect/go.mod it should say

```
require github.com/uptrace/bun v0.2.13
```

Currently it says v0.1.1 instead of v0.2.13

isgj:
fixes #56 
- Bump the required `bun` version to the one to be released
- Add only .mod files
- Commit before tagging

### To note
Even though this will add only the mod files to the stage, it will commit whatever is on stage. So should there be a check that fails immediately if stage is not empty? Or when committing default to `MESSAGE` if provided so with one commit you add changes and bump the version?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,15 @@ test:
 	done
 
 tag:
+	go mod edit -require github.com/uptrace/bun@v$(VERSION) dialect/pgdialect/go.mod
+	go mod edit -require github.com/uptrace/bun@v$(VERSION) dialect/mysqldialect/go.mod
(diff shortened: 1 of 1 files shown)


vmihailenco:
>So should there be a check that fails immediately if stage is not empty?

Yes, I think so. Probably `make tag` should be replaced with a shell script which is more flexible to work with. Merging for now.

PR #79, opened 2021-07-19
mysqldialect: allow use bool with tinyint

j2gg0s:
Do we have any plan to support convert between tinyint and bool.
Consider that MySQL always use tinyint to save bool value.

Feel free to reject to change this PR

### DIFF
--- a/dialect/mysqldialect/append.go
+++ b/dialect/mysqldialect/append.go
@@ -15,6 +15,10 @@ func appender(typ reflect.Type) schema.AppenderFunc {
 	if typ == timeType {
 		return appendTimeValue
 	}
+	switch typ.Kind() {
+	case reflect.Bool:
+		return appendBoolValue
+	}
 	return schema.Appender(typ, customAppender)
 }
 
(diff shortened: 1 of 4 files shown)


vmihailenco:
Thanks for the PR. I was not aware that boolean is an alias for tinyint in MySQL.

But it looks like this change does not improve anything except making the generated SQL look more mysqlish which is not much by itself. And without this change we have less code to support and potentially less bugs. I think we should not merge this until there are more substantial gains.

j2gg0s:
@vmihailenco this pr is just as example, and maybe not the best solution.

MySQL has not real BOOLEAN type.
```
mysql> CREATE TABLE t(b BOOLEAN);
Query OK, 0 rows affected (0.02 sec)

mysql> SHOW CREATE TABLE t\G
*************************** 1. row ***************************
       Table: t
Create Table: CREATE TABLE `t` (
  `b` tinyint(1) DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1
1 row in set (0.00 sec)
```

[bun](https://github.com/uptrace/bun) suppot auto convert int to bool [scanBool](https://github.com/uptrace/bun/blob/master/schema/scan.go#L94) .
[database/sql](https://github.com/golang/go) also support auto convert int to bool [ConvertValue](https://github.com/golang/go/blob/master/src/database/sql/driver/types.go#L63) .
So, auto convert int to bool maybe a common practice.

However, when use [go-sql-driver/mysql](https://github.com/go-sql-driver/mysql) without prepare statement, it will use MySQL's text protocol. Under this way, it will read TINYINT as `[]uint8`. So we need do extra work for this.

Another solution without only few changes:
```
diff --git a/schema/scan.go b/schema/scan.go
index f45ee01..edec442 100644
--- a/schema/scan.go
+++ b/schema/scan.go
@@ -102,6 +102,11 @@ func scanBool(dest reflect.Value, src interface{}) error {
        case int64:
                dest.SetBool(src != 0)
                return nil
+       case []byte:
+               if len(src) == 1 {
+                       dest.SetBool(src[0] != 48)
+                       return nil
+               }
        }
        return fmt.Errorf("bun: can't scan %#v into %s", src, dest.Type())
 }
```

vmihailenco:
>Another solution without only few changes:

I like it and we already do a similar thing for other types. I will send a PR.

PR #95, opened 2021-08-10
Add release.sh script

vmihailenco:
Fixes https://github.com/uptrace/bun/issues/56

@monkey92t @Codebreaker101 please take a look and let me know what do you think (RELEASING.md should have a brief outline). I plan to use the same script to release go-redis as well.

As the next step we can add https://github.com/softprops/action-gh-release to automatically make a GitHub release.


vmihailenco:
Merging since I need to make a release using these new tools ;)

PR #117, opened 2021-08-21
dbfixture: use yaml.Unmarshaler when available

vmihailenco:
@d-fal this should probably fix the problem with custom yaml.Unmarshaler. Could you try this code?

### DIFF
--- a/dbfixture/fixture.go
+++ b/dbfixture/fixture.go
@@ -179,7 +179,7 @@ func (f *Fixture) addRow(ctx context.Context, table *schema.Table, row row) erro
 			return err
 		}
 
-		if err := f.decodeField(strct, field, value); err != nil {
+		if err := f.decodeField(strct, field, &value); err != nil {
 			return fmt.Errorf("dbfixture: decoding %s failed: %w", key, err)
 		}
 	}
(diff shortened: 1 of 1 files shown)


dbfixture/fixture.go
```diff
@@ -233,6 +233,10 @@ func (f *Fixture) decodeField(strct reflect.Value, field *schema.Field, value ya
 		}
 	}
 
+	if v, ok := iface.(yaml.Unmarshaler); ok {
+		return v.UnmarshalYAML(value)
```
d-fal, line 237: the issue that's I am working on is still big number. yaml decoder, predicts the type for integers and treat the bigger ones as float. Inside UnmarshalYAML, one should even change the decoder prediction. 

```
var _ yaml.Unmarshaler = (*Bigint)(nil)

func (b *Bigint) UnmarshalYAML(value *yaml.Node) error {
	var err error
	type tmpBig Bigint

	if err := value.Decode((*tmpBig)(b)); err != nil {
		return err
	}

	b, err = FromString(value.Value)
	return err

}
```

Therer is a problem here, value.Kind expects an **yaml.ScalarNode** type, but the converted type is not. Here is the generated error:

>> cannot unmarshal !!str `1000` into bigint.tmpBig

vmihailenco: My guess is that it should be:

```go
var _ yaml.Unmarshaler = (*Bigint)(nil)

func (b *Bigint) UnmarshalYAML(value *yaml.Node) error {
	var str string
	if err := value.Decode(&str); err != nil {
		return err
	}

        var err error
	b, err = FromString(str)
	return err
}
```

But we could just ignore this problem if we can't solve it now... First we need to get the code in and iterate on it later.

d-fal: It soothes the panic, but the value is not being loaded, remains zero while it should be i.e. 100.

PR #119, opened 2021-08-24
bunbig added and the float time is here as well.

d-fal:
as @vmihailenco requested, extra/bunbig with deeper granularity is added.


vmihailenco:
Looks good, but please fix golangci-lint before merging

vmihailenco (approved):

d-fal:
readme updated with a proper example

vmihailenco:
Thanls :+1:

Issue #212, opened 2021-09-21
`panic: unaligned 64-bit atomic operation` running on 32 bit arm architecture


tsmethurst:
Hi! I'm trying to run GoToSocial on a Raspberry Pi Zero W, which uses 32 bit arm architecture. GoToSocial uses Bun for sqlite and postgres connections. Unfortunately, I run into the following panic when running gts on my pi:

```
panic: unaligned 64-bit atomic operation

goroutine 1 [running]:
runtime/internal/atomic.panicUnaligned()
	/usr/local/go/src/runtime/internal/atomic/unaligned.go:8 +0x24
runtime/internal/atomic.Xadd64(0x3082214, 0x1)
	/usr/local/go/src/runtime/internal/atomic/atomic_arm.s:256 +0x14
github.com/uptrace/bun.(*DB).beforeQuery(0x30821e0, {0x1be8ee0, 0x2c28608}, {0x1be6624, 0x30b11e0}, {0x3048000, 0xb5}, {0x0, 0x0, 0x0})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/uptrace/bun/hook.go:56 +0x34
github.com/uptrace/bun.(*baseQuery).exec(0x30b11e0, {0x1be8ee0, 0x2c28608}, {0x1be6624, 0x30b11e0}, {0x3048000, 0xb5})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/uptrace/bun/query_base.go:465 +0x64
github.com/uptrace/bun.(*CreateTableQuery).Exec(0x30b11e0, {0x1be8ee0, 0x2c28608}, {0x0, 0x0, 0x0})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/uptrace/bun/query_table_create.go:254 +0xf0
github.com/uptrace/bun/migrate.(*Migrator).Init(0x306b3b0, {0x1be8ee0, 0x2c28608})
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/vendor/github.com/uptrace/bun/migrate/migrator.go:87 +0x1a8
github.com/superseriousbusiness/gotosocial/internal/db/bundb.doMigration({0x1be8ee0, 0x2c28608}, 0x30821e0, 0x308ca40)
	/home/tobi/go/src/github.com/superseriousbusiness/gotosocial/internal/db/bundb/bundb.go:84 +0x90
[...]

PR #213, opened 2021-09-22
fix: change DBStats to use uint32 instead of uint64 to make it work o…

vmihailenco:
Fixes https://github.com/uptrace/bun/issues/212

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ test:
 	set -e; for dir in $(ALL_GO_MOD_DIRS); do \
 	  echo "go test in $${dir}"; \
 	  (cd "$${dir}" && \
-	    go test ./... && \
+	    go test && \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -4,7 +4,8 @@ test:
 	set -e; for dir in $(ALL_GO_MOD_DIRS); do \
 	  echo "go test in $${dir}"; \
 	  (cd "$${dir}" && \
-	    go test ./... && \
+	    go test && \
+	    env GOOS=linux GOARCH=386 go test && \
```
NyaaaWhatsUpDoc, line 8: it's specifically 32bit ARM and 32bit MIPS where those previous issues fail, so depending on host platform this test may not actually trigger the previous issue

vmihailenco: Got it, but it worked fine in this case.

PR #350, opened 2021-12-07
change the default logoutput to stderr

### LINKED ISSUE #349: bundebug should print to stderr instead of stdout
it's a rather trivial change,
should i PR?

https://github.com/uptrace/bun/blob/47a60c16988cc3b072022bcb01b9892df05f72c9/extra/bundebug/debug.go#L104

aep:
also add an option to set it to any writer

fixes #349


vmihailenco:
Thanks

PR #352, opened 2021-12-08
feat: Add `Dialect()` to Transaction and IDB interface.

elliotcourant:
This is just a little addition to better support dialect specific code within transactions or functions that just want an IDB input.


vmihailenco:
Thanks

PR #355, opened 2021-12-11
A bit more detailed benchmark section

gandaldf:


vmihailenco:
Thanks. One thing that I've noticed is that XORM does not use transactions to insert data while most other clients (including Bun) do use transactions.

Also GORM/go-pg are outdated...

Issue #30, opened 2021-06-22
Map embedded structures to fields


jpascal:
I think this feature will be nice to have

```go
type Permissions struct {
	View   bool
	Create bool
	Update bool
	Delete bool
}

type Role struct {
	Enable   bool
	Name     string
	Users    Permissions `bun:"embedded;embeddedPrefix:users_"`
	Profiles Permissions `bun:"embedded;embeddedPrefix:profiles_"`
	Roles    Permissions `bun:"embedded;embeddedPrefix:roles_"`
}
```

vmihailenco:
Makes sense. But the tag could look like `embed:users_` instead. I will try to implement this for v0.3.

imraan-go:
May I know what does this accomplish?

vmihailenco:
The result will be table like this

```sql
create table roles (
    users_view,
    users_create,
    users_update,
    users_delete,

    profiles_view,
    profiles_create,
    profiles_update,
    profiles_delete,

    roles_view,
    roles_create,
    roles_update,
    roles_delete
)
```

It is something like a [composite type](https://www.postgresql.org/docs/13/rowtypes.html) in PostgreSQL.

jestradaramos:
Is this still expected to be implemented? Seems like a useful feature.

PR #361, opened 2021-12-16
feat: add model embedding via embed:prefix_

vmihailenco:
Fixes https://github.com/uptrace/bun/issues/30

This requires more testing and docs.


sbward:
Works in my app 👍

frederikhors:
@vmihailenco is there any docs about this?

vmihailenco:
There is an example at https://bun.uptrace.dev/guide/models.html#embedding

PR #362, opened 2021-12-16
example: add fixture example for XID (not working)

smyrman:


vmihailenco:
This should be working on master branch (unreleased version). Please re-open if that is not the case.

PR #365, opened 2021-12-18
feat: use one `IDB` instead of `*DB` with `IConn`

### LINKED ISSUE #364: Select with `HasMany` and `ManyToMany` does not use current transaction
When performing a join on a `HasMany` or `ManyToMany` relation, bun starts a new select query as such:

https://github.com/uptrace/bun/blob/25ac1477f98451c0974dc3b57a5af1e91b863676/query_select.go#L346-L349

Unfortunately creating the new query via `q.db` ignores whether `q.conn` might be a `*sql.Tx`, which can break code that performs this join during a transaction. Using `q.db` creates a new connection to perform the query, which will break queries that select data created or modified during a transaction.

sbward:
Fixes #364 

Instead of holding a `*DB` and an `IConn` in each Query struct, we can have `DB`, `Conn`, and `Tx` all implement `IDB` and embed `*DB` so that its methods are promoted to `Conn` and `Tx` by default. This allows query logic to ignore whether the current `IDB` is a transaction or not.

This approach also has the advantage of preventing issues like #364 (which is fixed by this PR) by encapsulating the type of connection.

## Context and Hooks

(This part is subject to change. I might be able to refactor it out.)

Unifying `*DB` and `IConn` changes which hooks are invoked and when. To fix this issue I've added a `scopeContext` that can carry the `Query`, `Model`, and `*QueryEvent` associated with the current Context. This allows us to remove many of the hook invocations since we can dynamically check the `scopeContext` from `IDB` implementations to decide whether or not to invoke hooks, so that before and after hooks are called exactly one time per query.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,53 @@
+package bun
+
+import (
+	"context"
+)
+
+type contextKey struct{}
+
+type scope struct {
+	query Query
+	model Model
+	event *QueryEvent
+}
+
+var scopeKey = contextKey{}
+
+func (s *scope) copy() *scope {
+	return &scope{
+		query: s.query,
+		model: s.model,
+		event: s.event,
+	}
+}
+
+func scopeFromContext(ctx context.Context) *scope {
+	state := ctx.Value(scopeKey)
+	if state == nil {
+		return &scope{}
+	}
+	return state.(*scope)
+}
+
+func withQueryScope(ctx context.Context, q Query) context.Context {
+	scopeVal := scopeFromContext(ctx).copy()
+	scopeVal.query = q
+	return context.WithValue(ctx, scopeKey, scopeVal)
+}
+
+func withModel(ctx context.Context, m Model) context.Context {
+	scopeVal := scopeFromContext(ctx).copy()
+	scopeVal.model = m
+	return context.WithValue(ctx, scopeKey, scopeVal)
+}
+
+func withEvent(ctx context.Context, e *QueryEvent) context.Context {
+	scopeVal := scopeFromContext(ctx).copy()
+	scopeVal.event = e
+	return context.WithValue(ctx, scopeKey, scopeVal)
+}
+
+func isManagedQuery(ctx context.Context) bool {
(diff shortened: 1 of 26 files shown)


query_base.go
```diff
@@ -57,17 +45,29 @@ type IDB interface {
...
+	Table(reflect.Type) *schema.Table
+
+	_flags() internal.Flag
+	makeQueryBytes() []byte
+
+	afterQuery(
+		ctx context.Context,
+		res sql.Result,
+		err error,
+	)
```
vmihailenco, line 60: This was a big interface before and now it grew even bigger and got some private methods. Not nice :)

context.go
```diff
@@ -0,0 +1,53 @@
...
+
+func withEvent(ctx context.Context, e *QueryEvent) context.Context {
+	scopeVal := scopeFromContext(ctx)
+	scopeVal.event = e
+	return context.WithValue(ctx, scopeKey, scopeVal.copy())
+}
+
+func isManagedQuery(ctx context.Context) bool {
+	return scopeFromContext(ctx).query != nil
+}
```
vmihailenco, line 53: I am not an anti-context purist and don't mind using it to solve something quickly, but context.Context should not be put in the core of the design. We may need something like scope some day, but I don't think it be done via `context.Context`...

vmihailenco:
Thanks for the PR. I like the intention to unify everything under the common interface / context / scope, but this changes introduces new issues:
- Every method added to `*DB` becomes automatically available in `Tx` and `Conn` which may not be desired and requires additional care/work.
- Most calls should go through `context.Context` and `IDB` interface. That is not going to be fast. It probably does not matter in practice, but none likes slow code. Also, private methods in `IDB` interface.
- Overall It is slightly more complex and harder to understand.

---

I think we should keep existing design and fix https://github.com/uptrace/bun/issues/364 with minimal changes. The existing design (or lack of it) has some flaws, but they are simple to understand and cope with.

sbward:
@vmihailenco Thanks for your feedback, I've submitted a much simpler fix here: #366

PR #366, opened 2021-12-18
fix: reuse tx for relation join

### LINKED ISSUE #364: Select with `HasMany` and `ManyToMany` does not use current transaction

sbward:
Fixes #364

- Minimal change for reusing an existing transaction when joining relations.
- (Minor) Fixes `IDB`/`IConn` interface checking and comments.

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -37,10 +37,10 @@ var (
 	_ IConn = (*sql.Tx)(nil)
 	_ IConn = (*DB)(nil)
 	_ IConn = (*Conn)(nil)
-	_ IConn = (*Tx)(nil)
+	_ IConn = (Tx{})
 )
 
-// IDB is a common interface for *bun.DB, bun.Conn, and bun.Tx.
+// IDB is a common interface for *bun.DB, and bun.Tx.
 type IDB interface {
 	IConn
 	Dialect() schema.Dialect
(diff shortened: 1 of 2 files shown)


query_select.go
```diff
@@ -344,9 +344,9 @@ func (q *SelectQuery) selectJoins(ctx context.Context, joins []relationJoin) err
 		case schema.HasOneRelation, schema.BelongsToRelation:
 			err = q.selectJoins(ctx, j.JoinModel.getJoins())
 		case schema.HasManyRelation:
-			err = j.selectMany(ctx, q.db.NewSelect())
+			err = j.selectMany(ctx, q.idb(ctx).NewSelect())
 		case schema.ManyToManyRelation:
-			err = j.selectM2M(ctx, q.db.NewSelect())
+			err = j.selectM2M(ctx, q.idb(ctx).NewSelect())
```
vmihailenco, line 349: What about `q.db.NewSelect().Conn(q.conn)`?

query_base.go
```diff
@@ -60,9 +60,8 @@ type IDB interface {
 }
 
 var (
-	_ IConn = (*DB)(nil)
-	_ IConn = (*Conn)(nil)
-	_ IConn = (*Tx)(nil)
+	_ IDB = (*DB)(nil)
+	_ IDB = Tx{}
```
vmihailenco, line 64: This used to work before. Probably should be reverted.

query_select.go
sbward: 👍

PR #373, opened 2021-12-21
fix: append where to index create

### LINKED ISSUE #369: Postgres NewCreateIndex not adding "Where" to Where statement
not adding "Where" to query when using CreateNewIndex:

```
  query.DB().NewCreateIndex().
    Unique().Model((*User)(nil)).
    Where("\"deleted_at\" IS NULL").
    Index("users_email_key").
    Column("email").
    Exec(ctx)
```
Creates This:

`
CREATE UNIQUE INDEX "users_email_key" ON "users" ("email")("deleted_at" IS NULL)
`

jelwan:
fixes #369


vmihailenco:
Thanks :+1:

PR #381, opened 2021-12-25
feat: add anonymous fields with type name

sbward:
Enables support for anonymous fields with non-struct types.

### Example

```go
type MyString string

type A struct{
    MyString // Will create a field called `my_string` with VARCHAR/TEXT type
}
```

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -1087,3 +1087,26 @@ func testEmbedModelPointer(t *testing.T, db *bun.DB) {
 	require.NoError(t, err)
 	require.Equal(t, *m1, m2)
 }
+
+func testEmbedTypeField(t *testing.T, db *bun.DB) {
+	type Embed string
+	type Model struct {
+		Embed
+	}
+
+	ctx := context.Background()
+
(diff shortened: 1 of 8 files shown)


vmihailenco:
Is this any better than:

```go
type A struct{
    MyString MyString
}
```

This way it is less code / tests / docs to write and support.

sbward:
My situation is that I want to embed type so that its methods are promoted. If I don't embed it, then I need to manually write each of the methods I want to promote. So in a way, the burden of writing more code and tests is now pushed onto the developer.

I will be happy to add tests and docs to this PR if it will help it get approved :)

vmihailenco:
Looks like encoding/json supports patterns like this so we probably should too.

Please add some tests and maybe comments to the code, because I can't spot any changes in the diff :) I would add a `InsertQuery` test to [query_test.go](https://github.com/uptrace/bun/blob/master/internal/dbtest/query_test.go) and more functional test to [db_test.go](https://github.com/uptrace/bun/blob/master/internal/dbtest/db_test.go).

sbward:
Updated with tests and comments.

vmihailenco:
Looks good, thanks. I will merge this tomorrow.

PR #384, opened 2021-12-27
fix: unique module name for opentelemetry example

sbward:
Update the opentelemetry example so that it has a unique module name. Currently this example has the same name as the basic example (`github.com/uptrace/bun/example/basic`) which creates a namespace collision and may break some tools, such as `gopls` in experimental module mode.


vmihailenco:
Thanks

PR #388, opened 2021-12-30
build: update `go` directive in `go.mod` to 1.17

Juneezee:
This PR updates the `go` directive in `go.mod` file by running `go mod tidy -go=1.17` to enable [module graph pruning](https://golang.org/ref/mod#graph-pruning) and [lazy module loading](https://golang.org/ref/mod#lazy-loading).

**Note 1:** This does not prevent users with earlier Go versions from successfully building packages from this module.

**Note 2:** The additional `require` directive is used to record indirect dependencies for Go 1.17 or higher, see https://go.dev/ref/mod#go-mod-file-go.

### DIFF
--- a/dbfixture/go.mod
+++ b/dbfixture/go.mod
@@ -1,11 +1,18 @@
 module github.com/uptrace/bun/dbfixture
 
-go 1.16
+go 1.17
 
 replace github.com/uptrace/bun => ../
 
 require (
 	github.com/uptrace/bun v1.0.20
-	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e // indirect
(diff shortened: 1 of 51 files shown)


vmihailenco:
Thanks, completely missed those changes. Is there a command to convert go.mod files or you just remove the old one and run `go mod tidy`?

Juneezee:
> Thanks, completely missed those changes. Is there a command to convert go.mod files or you just remove the old one and run `go mod tidy`?

@vmihailenco Hi! Thanks for accepting the PR. I ran `go mod tidy -go=1.17` to update the `go` directive (as documented in here: https://go.dev/ref/mod#go-mod-tidy).

I believe manually changing from `go 1.16` to `go 1.17` in `go.mod` and then run `go mod tidy` would work too, but this method is slightly more tedious.

PR #392, opened 2022-01-04
feat: create new queries from baseQuery

sbward:
As suggested by https://github.com/uptrace/bun/pull/390#issuecomment-1004551718

Allow users to conveniently create nested queries from `*baseQuery` similarly to `IDB` values.


vmihailenco:
Thanks :+1:

PR #404, opened 2022-01-18
chore: add identity feature for mssql

vmihailenco:


dialect/mysqldialect/dialect.go
```diff
@@ -29,6 +29,7 @@ func New() *Dialect {
 	d := new(Dialect)
 	d.tables = schema.NewTables(d)
 	d.features = feature.AutoIncrement |
+		feature.Identity | // for mssql dialect
```
vmihailenco, line 32: @gandaldf this should be in mssql dialect

gandaldf:
@vmihailenco This is more or less the same code I've wrote, my question is about field tags: we should tag them as "auto_increment" even if they are "identity"?
I'm new and I would like to understand the general philosophy of the project.
Soon I will publish my code, it's just a draft for now... nothing reliable.

vmihailenco:
@gandaldf it is all about convenience. If you find a cleaner way to implement this - go ahead with it. But it might be a good idea to discuss the new way before going full in :)

Ultimately, we need some boolean to switch between "AUTO_INCREMENT" and "IDENTITY" in different dialects. `feature.Has` looks like a sensible way to do it.

>Soon I will publish my code, it's just a draft for now... nothing reliable.

Don't hesitate to ping me if you are stuck :+1:

vmihailenco:
>I'm new and I would like to understand the general philosophy of the project.

Regarding philosophy... the goal is to find a common ground and features between Postgres/MySQL/MSSQL and provide uniform way to define models. So it must be the same option name to enable auto increments in different databases if they support the option, not different "branded" option names in different dialects.

So using `bun:",pk,autoincrement"` in all dialects is the right way forward if `identity` is just a different name for autoincrement.

PR #409, opened 2022-01-18
feat: add IndexHash to feature flags

mystic123:
This adds IndexHash feature to indicate DB engine supports HASH indexes (CREATE INDEX ... USING HASH ...)
e.g. PG and MySQL both support this, while SQLite does not.


vmihailenco:
@mystic123 what is the use case for this?

Before this PR, SQLite users will get some SQL syntax error. With this PR, they get `ErrUnsupportedFeature` which is rather obscure error message. Why not to let it fail with a syntax error?

PR #416, opened 2022-01-24
feat(db): support control db's features

j2gg0s:
Let's add functions to support enable/disable db's features by user.

In my case, i need disable `InsertReturning`.

### DIFF
--- a/db.go
+++ b/db.go
@@ -30,6 +30,22 @@ func WithDiscardUnknownColumns() DBOption {
 	}
 }
 
+func EnableFeature(features ...feature.Feature) DBOption {
+	return func(db *DB) {
+		for _, feature := range features {
+			db.features |= feature
+		}
+	}
+}
+
+func DisableFeature(features ...feature.Feature) DBOption {
+	return func(db *DB) {
+		for _, feature := range features {
+			db.features &= ^feature
+		}
+	}
+}
+
 type DB struct {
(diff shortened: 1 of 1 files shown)


vmihailenco:
`feature` is a low-level API so use at your own risk. Bun does not guarantee it won't change.

Do you need this per-query basic or this should be done once? I would prefer something like

```go
db := bun.NewDB(db, dialect.WithoutFeature(feature.InsertReturning))
```

j2gg0s:
Yes, `feature` is low-level API. But i think support functions to somebody need is not a bad choice.

I rename function's name.

vmihailenco:
@j2gg0s what do you think about making this a method on a `Dialect`? The more accurate example is:

```go
sqldb, err := sql.Open("mysql", "root:pass@/test")
if err != nil {
	panic(err)
}

db := bun.NewDB(sqldb, mysqldialect.New().WithoutFeature(feature.InsertReturning))
```

Yes, we need to repeat for all 3 dialects, but the API is more cleaner.

j2gg0s:
Sorry, i did not notice that yout want to add this function to dialect.

However, if we think this may be not very correct, let's add this function when we have to?
I have deal with my special case with other solution.

vmihailenco:
Okay, sounds like you've found another solution for your problem. Let's close this then?

PR #417, opened 2022-01-24
feat(insert): get last insert id only with pk support auto increment

j2gg0s:
Should we try to get last insert id only with table's pk support auto incrment?
This may make incompatible upgrade!

Or can we support some option to user?

In my case, i use string as pk.


vmihailenco:
This looks like a reasonable safe change. I guess it can break some code, but it should not stop us.

vmihailenco (approved):

Issue #197, opened 2021-09-14
Add CreateTableQuery.DetectForeignKeys or something similar


vmihailenco:

vmihailenco:
Closed thanks to @iamgoroot

PR #430, opened 2022-01-30
Add CreateTableQuery.DetectForeignKeys

iamgoroot:
Add CreateTableQuery.DetectForeignKeys method
https://github.com/uptrace/bun/issues/197


vmihailenco:
Thanks :+1: It would be nice if we can detect more foreign keys by exploring relations from the other side, but this is a good start.

frederikhors:
Can we add this method to the docs?

PR #440, opened 2022-02-08
feat: Make soft deletes work with mssql 

iamgoroot:
fixes fot mssql soft_delete and bulk update
https://github.com/uptrace/bun/issues/429

### DIFF
--- a/dialect/feature/feature.go
+++ b/dialect/feature/feature.go
@@ -28,4 +28,5 @@ const (
 	TableNotExists
 	OffsetFetch
 	SelectExists
+	UpdateAliasAsMultiTable
 )
--- a/dialect/feature/feature.go
+++ b/dialect/feature/feature.go
@@ -28,5 +28,5 @@ const (
 	TableNotExists
 	OffsetFetch
 	SelectExists
-	UpdateAliasAsMultiTable
+	UpdateFromTable
 )
--- a/dialect/mssqldialect/dialect.go
+++ b/dialect/mssqldialect/dialect.go
@@ -45,7 +45,8 @@ func New() *Dialect {
 		feature.DefaultPlaceholder |
 		feature.Identity |
 		feature.Output |
-		feature.OffsetFetch
+		feature.OffsetFetch |
+		feature.UpdateAliasAsMultiTable
 	return d
 }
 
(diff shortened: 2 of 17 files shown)


vmihailenco:
What do you think about renaming `UpdateAliasAsMultiTable` to `UpdateFromTable`? I can't understand the logic behind `UpdateAliasAsMultiTable` name... Then this PR should be ready for merging.

Overall this looks like a nice addition for mssql driver, but we still need to change soft-deletes to generate simpler queries, e.g. instead of 

```sql
UPDATE "soft_deletes" SET "deleted_at" = [TIME] FROM "soft_deletes" AS "soft_delete" WHERE "soft_delete"."deleted_at" IS NULL AND ("soft_delete"."id" = NULL)
```

should be

```sql
UPDATE "soft_deletes" SET "deleted_at" = [TIME] WHERE "soft_deletes"."deleted_at" IS NULL AND ("soft_deletes"."id" = NULL)
```

Probably a task for a separate PR.

iamgoroot:
Done. Hopefully would not be confused with  `update from select ` statement
Other things that could be adressed:
* There are probably other features in mssql that could be enabled.
* `unique` doesnt work with soft delete (any db actually). It is documented but I haven't found an issue. have test. can probably look into it
* benchmark performance is quite poor in comparison

> What do you think about renaming `UpdateAliasAsMultiTable` to `UpdateFromTable`? I can't understand the logic behind `UpdateAliasAsMultiTable` name... Then this PR should be ready for merging.
> 
> Overall this looks like a nice addition for mssql driver, but we still need to change soft-deletes to generate simpler queries, e.g. instead of
> 
> ```sql
> UPDATE "soft_deletes" SET "deleted_at" = [TIME] FROM "soft_deletes" AS "soft_delete" WHERE "soft_delete"."deleted_at" IS NULL AND ("soft_delete"."id" = NULL)
> ```
> 
> should be
> 
> ```sql
> UPDATE "soft_deletes" SET "deleted_at" = [TIME] WHERE "soft_deletes"."deleted_at" IS NULL AND ("soft_deletes"."id" = NULL)
> ```
> 
> Probably a task for a separate PR.

vmihailenco:
>unique doesnt work with soft delete (any db actually). It is documented but I haven't found an issue. have test. can probably look into it

Let's add the test even if it does not pass. `t.Skip` is fine.

I guess you've seen https://bun.uptrace.dev/guide/soft-deletes.html#unique-indexes already

>benchmark performance is quite poor in comparison

Again, let's add the benchmark even if results are not impressive. It is good to know the number anyway.

PR #443, opened 2022-02-12
feat: add soft_delete partial index creation to unsure "unique"

iamgoroot:
* Should be enabled for ms-sql as well 
https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver15#syntax 
* mysql is tricky. 
Common approach is to use `CREATE UNIQUE INDEX team_members_role_team_unique_key ON team_members (role, team, active) ` where `active` is bool DEFAULT 1 that can be used instead of actual safe_delete to unsure uniqueness
Obviously field `active` should be updated along with soft_delete field for that

### DIFF
--- a/query_table_create.go
+++ b/query_table_create.go
@@ -3,6 +3,7 @@ package bun
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"sort"
 	"strconv"
 
@@ -22,6 +23,7 @@ type CreateTableQuery struct {
 	fks         []schema.QueryWithArgs
 	partitionBy schema.QueryWithArgs
 	tablespace  schema.QueryWithArgs
+	indexes     []*CreateIndexQuery
 }
 
 var _ Query = (*CreateTableQuery)(nil)
@@ -227,8 +229,21 @@ func (q *CreateTableQuery) appendUniqueConstraints(fmter schema.Formatter, b []b
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
-
 	for _, key := range keys {
+		if q.table.SoftDeleteField != nil && q.Dialect().Features().Has(feature.PartialIndex)  {
+			keyName := fmt.Sprintf("%s_%s_unique_key", q.table.Name, key)
+			index := q.NewCreateIndex().
+				Index(keyName).
+				TableExpr(string(q.table.SQLName)).
+				Unique()
+			for _, field := range unique[key] {
+				index.Column(string(field.Name))
+			}
+			where := fmter.AppendQuery([]byte(q.table.SoftDeleteField.SQLName), " IS NULL")
+			index.Where(string(where))
+			q.indexes = append(q.indexes, index)
+			continue
+		}
 		if key == "" {
 			for _, field := range unique[key] {
 				b = q.appendUniqueConstraint(fmter, b, key, field)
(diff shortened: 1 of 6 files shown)


vmihailenco:
I think we should document the available options at [docs](https://bun.uptrace.dev/guide/soft-deletes.html#unique-indexes) and let users chose their medicine. I always opposed to creating indexes in Bun models and the task looks even more daunting when we start considering soft deletes and 4 different DBMS.

I would leave creating indexes to the users.

iamgoroot:
Perhaps it could be implemented as separate method like [WithForeignKeys](https://github.com/uptrace/bun/commit/8333021634c903cd9b84ad7a218a4b0968c7baf4) so user can setup effortlessly his indexes?


```
If your DBMS does not allow to use expressions in indexed columns, you can configure Bun to not append zero time as NULL using allowzero option

-	 DeletedAt time.Time `bun:",soft_delete"`
+	 DeletedAt time.Time `bun:",soft_delete,allowzero"`
```
Would be nice having same model sources for all databases. 
Is it sensible to enable allowzero for soft delete fields on dialects that don't support feature.PartialIndex?

vmihailenco:
The main problem here is that we only support this via `CreateTable` which is not flexible at all when you already have a table. Perhaps we should add something like GORM's AutoMigrate and then this feature would be more useful.

For now let's add this under some flag like WithForeignKey so users can opt-out.

>Is it sensible to enable allowzero for soft delete fields on dialects that don't support feature.PartialIndex?

I plan to release v1.0.x that will require users to enable `,pk` and `,nullzero` explicitly. So `,nopk` and `,allowzero` will go away in v1.1.0+

query_table_create.go
```diff
@@ -227,8 +229,21 @@ func (q *CreateTableQuery) appendUniqueConstraints(fmter schema.Formatter, b []b
...
+			keyName := fmt.Sprintf("%s_%s_unique_key", q.table.Name, key)
+			index := q.NewCreateIndex().
+				Index(keyName).
+				TableExpr(string(q.table.SQLName)).
+				Unique()
+			for _, field := range unique[key] {
+				index.Column(string(field.Name))
+			}
+			where := fmter.AppendQuery([]byte(q.table.SoftDeleteField.SQLName), " IS NULL")
+			index.Where(string(where))
```
vmihailenco, line 243: Should be `index.Where("? IS NULL", ch.Safe(table.SoftDeleteField.SQLName))`

query_table_create.go
```diff
@@ -227,8 +229,21 @@ func (q *CreateTableQuery) appendUniqueConstraints(fmter schema.Formatter, b []b
 		keys = append(keys, key)
 	}
 	sort.Strings(keys)
-
 	for _, key := range keys {
+		if q.table.SoftDeleteField != nil && q.Dialect().Features().Has(feature.PartialIndex)  {
+			keyName := fmt.Sprintf("%s_%s_unique_key", q.table.Name, key)
+			index := q.NewCreateIndex().
+				Index(keyName).
```
vmihailenco, line 236: We somehow should chose naming schema for these indexes so we can check the presence of the index by name in the hypothetical `AutoMigrate` method.

vmihailenco:
Closing since there are conflicts.

Issue #462, opened 2022-02-25
func (*bun.CreateTableQuery) WithForeignKeys() results in unwanted foreign keys when m2m relationships are in place


brueli:
Affected Versions: bun @ 1.0.25

### Reproduction
```
package test

import (
	"context"
	"log"
	"time"

	"github.com/uptrace/bun"
)

type Profile struct {
	ID     int64     `bun:",pk,autoincrement"`
	Policy []*Policy `bun:"m2m:policy_to_profiles,join:Profile=Policy"`
}

type Policy struct {
	ID      int64     `bun:",pk,autoincrement"`
	Profile []*Profile `bun:"m2m:policy_to_profiles,join:Policy=Profile"`
}

type PolicyToProfile struct {
	ProfileID int64    `bun:",pk"`
	PolicyID  int64    `bun:",pk"`
	Profile   *Profile `bun:"rel:belongs-to,join:profile_id=id"`
	Policy    *Policy  `bun:"rel:belongs-to,join:policy_id=id"`
}

func DeployModel(db *bun.DB) {
	m2mModels := []interface{}{
		(*PolicyToProfile)(nil),
	}

	tableDefs := []interface{}{
		(*Profile)(nil),
		(*Policy)(nil),
		(*PolicyToProfile)(nil),
	}

	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
	defer cancel()

	// drop all tables
	db.NewDropTable().Model(tableDefs[2]).IfExists().Exec(ctx)
	db.NewDropTable().Model(tableDefs[1]).IfExists().Exec(ctx)
	db.NewDropTable().Model(tableDefs[0]).IfExists().Exec(ctx)

	// register many-2-many models
	for _, m2mModel := range m2mModels {
		db.RegisterModel(m2mModel)
	}

	// create the tables with IfNotExists() and WithForeignKeys() options
	for key, tableDef := range tableDefs {
		if _, err := db.NewCreateTable().Model(tableDef).IfNotExists().WithForeignKeys().Exec(ctx); err != nil {
			log.Printf("cannot create table %d: %s", key, err)
			break
		}
	}

}
```

This results in the following error:
```
DROP TABLE IF EXISTS "policy_to_profiles"
DROP TABLE IF EXISTS "policies"
DROP TABLE IF EXISTS "profiles"
[...]

PR #463, opened 2022-02-25
Fix for bun issue #462

brueli:
Fix for https://github.com/uptrace/bun/issues/462


brueli:
need to fix commit message

PR #506, opened 2022-04-05
feat: add ApplyQueryBuilder

vmihailenco:
This allows to write code like this:

```go
func addWhere(q bun.QueryBuilder) QueryBuilder {
    return q.Where("id = ?", 123)
}

// Alternatively.

db.NewSelect().ApplyQueryBuilder(addWhere).Scan(ctx)
db.NewUpdate().ApplyQueryBuilder(addWhere).Exec(ctx)
db.NewDelete().ApplyQueryBuilder(addWhere).Exec(ctx)
```

### DIFF
--- a/query_delete.go
+++ b/query_delete.go
@@ -293,11 +293,17 @@ func (q *DeleteQuery) QueryBuilder() QueryBuilder {
 	return &deleteQueryBuilder{q}
 }
 
+func (q *DeleteQuery) ApplyQueryBuilder(fn func(QueryBuilder) QueryBuilder) *DeleteQuery {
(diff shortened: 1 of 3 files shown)


vmihailenco:
@derkan what do you think? Please note that I also renamed `Query() QueryBuilder` to `QueryBuilder() QueryBuilder`.

derkan:
@vmihailenco it is fine for me, this gives a better API for the lib.

YoungiiJC:
@vmihailenco could you please provide a real example of how this is used? I am struggling to figure out how to chain queries with this. thank you :)

PR #533, opened 2022-05-13
feat: add support for ON UPDATE and ON DELETE rules on belongs-to relationships from struct tags

antipopp:
I was missing this feature from `go-pg` and wanted to add it to the codebase. AFAIK `belongs-to` is the only rel that would benefit from it, as it's the only one that generates a column on the DB. But I still don't know this repo as deeply as I want to, so feel free to correct me. 

```golang
type User struct {
		ID   int    `bun:",pk,autoincrement"`
		Type string `bun:",pk"`
		Name string
}
type Deck struct {
		ID       int `bun:",pk,autoincrement"`
		UserID   int
		UserType string
		User     *User `bun:"rel:belongs-to,join:user_id=id,join:user_type=type,on_update:cascade,on_delete:set null"`
}
```

Also, should I tag the release for this?

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -254,6 +254,7 @@ func TestDB(t *testing.T) {
 		{testJSONValuer},
 		{testSelectBool},
 		{testFKViolation},
+		{testWithForeignKeysAndRules},
 		{testWithForeignKeys},
 		{testInterfaceAny},
 		{testInterfaceJSON},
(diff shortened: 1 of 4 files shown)


tinyfluffs:
Might be a good idea if we could have a unit test for this in db_test, similar to how foreign key violations are checked

[db_test.go#L820=](https://github.com/uptrace/bun/blob/af9dd3595eef2cebe47a1ab945e0b01e41ed1c24/internal/dbtest/db_test.go#L820=)

antipopp:
> Might be a good idea if we could have a test for this in db_test, similar to how foreign key violations are checked: [db_test.go#L820](https://github.com/uptrace/bun/blob/af9dd3595eef2cebe47a1ab945e0b01e41ed1c24/internal/dbtest/db_test.go#L820=)
> 
> Love the change though, been missing this feature myself

I've been thinking about the tests. What would a pattern look like? Create the table, insert entries, delete/update one and check that the FK has been updated?

also help me with commitlint lol

tinyfluffs:
Yeah, I wouldn't imagine it has to be anything too complex. Try and force an update/delete and check for the intended result.

Re commit lint, it's expecting this one's message to be formatted correctly:
[fix OnUpdate default string](https://github.com/uptrace/bun/pull/533/commits/15454387f9423a5c1512910f209c308c35bc9426)

You can do a rebase-edit to change the message - `fix: change OnUpdate default string` should be valid

internal/dbtest/db_test.go
```diff
@@ -879,7 +879,7 @@ func testWithForeignKeys(t *testing.T, db *bun.DB) {
 		ID       int `bun:",pk,autoincrement"`
 		UserID   int
 		UserType string
-		User     *User `bun:"rel:belongs-to,join:user_id=id,join:user_type=type"`
```
tinyfluffs, line 882: It'd still be good to test behavior without the new tags. Replacing this with additional tests as discussed

antipopp:
I've added the tests, tell me if there is anything else to fix

tinyfluffs:
LGTM

vmihailenco:
@antipopp thanks for the PR and @tinyfluffs thanks for reviewing :+1:

Issue #546, opened 2022-05-25
[bunotel] option to always use unformatted query


matoous:
Hi 👋 

I am super happy with this ORM so far and it suites are needs pretty well. My only concern/quirk so far is with the tracing. I would prefer the traces _not_ to contain the args (inlined) so we don't accidentally leak anything into the traces that shouldn't be there.

Would you be open for a PR that would add option to the `bunotel` to always keep the query unformatted?

PR #547, opened 2022-05-26
feat(bunotel): add option to enable formatting of queries

matoous:
Disable formatting by default (breaing change)
and add bunotel option that enables formatting.

Formatting has potential to leak information into the traces
as such not formatting the query by default seems like
safer option.

Fixes: https://github.com/uptrace/bun/issues/546

### DIFF
--- a/extra/bunotel/option.go
+++ b/extra/bunotel/option.go
@@ -20,3 +20,13 @@ func WithDBName(name string) Option {
 		h.attrs = append(h.attrs, semconv.DBNameKey.String(name))
 	}
 }
+
+// WithFormatting enables formatting of the query that is added
+// as the statement attribute to the trace.
+// This means that all placeholders and arguments will be filled first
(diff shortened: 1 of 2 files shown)


matoous:
@vmihailenco would you take a look please? This is breaking change, so if you want the default to stay as it is I will have to invert the logic and have option `WithUnfromattedQuery` or maybe `WithFormatting` that would take boolean/flag param.

vmihailenco:
@matoous thanks for the PR.

Let's change `WithFormatting` to accept a boolean flag - I have some reasoning [here](https://blog.uptrace.dev/posts/go-functional-options-named-args.html#bool-options). I am fine with having it disabled by default.

Also, what about renaming `WithFormatting` to `WithFormattedQuery`?

matoous:
@vmihailenco thanks for feedback, updated.

vmihailenco:
Thanks :+1:

PR #548, opened 2022-05-27
feat(config.go): add sslrootcert support to DSN parameters

adriansmares:
This PR adds support for `sslrootcert`, as implemented by [`libpq`](https://www.postgresql.org/docs/current/libpq-ssl.html).

### DIFF
--- a/driver/pgdriver/config.go
+++ b/driver/pgdriver/config.go
@@ -3,8 +3,10 @@ package pgdriver
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


driver/pgdriver/config.go
```diff
@@ -246,15 +248,66 @@ func parseDSN(dsn string) ([]Option, error) {
...
-	case "disable":
-		opts = append(opts, WithInsecure(true))
-	default:
-		return nil, fmt.Errorf("pgdriver: sslmode '%s' is not supported", sslMode)
+	if sslMode, sslRootCert := q.string("sslmode"), q.string("sslrootcert"); sslMode != "" || sslRootCert != "" {
+		tlsConfig := &tls.Config{}
+		switch sslMode {
+		case "disable":
+			tlsConfig = nil
+		case "allow", "prefer", "":
```
adriansmares, line 256: `allow` and `prefer` were not, and are still not entirely correct.

Specifically, if `allow` is used, the initial connection attempt should be done without TLS, and only if that fails, a TLS connection should be attempted.
If `prefer` is used, the reverse should happen (first with TLS, then without).

As the current semantics of `Config` doesn't really allow for these fallback semantics, I've avoided refactoring more stuff in this PR.

driver/pgdriver/proto.go
```diff
@@ -121,7 +121,11 @@ func enableSSL(ctx context.Context, cn *Conn, tlsConf *tls.Config) error {
 		return errors.New("pgdriver: SSL is not enabled on the server")
 	}
 
-	cn.netConn = tls.Client(cn.netConn, tlsConf)
+	tlsCN := tls.Client(cn.netConn, tlsConf)
+	if err := tlsCN.HandshakeContext(ctx); err != nil {
```
adriansmares, line 125: The handshake error would be hidden behind an `driver.ErrBadConn` otherwise. At least in this way the handshake error is returned. during the connection.

driver/pgdriver/config.go
vmihailenco: Thanks for the information - I was not aware of that... I guess we can discuss this in detail later if there is interest in making this work.

driver/pgdriver/proto.go
vmihailenco: Thanks for fixing this :+1:

vmihailenco (approved):
Looks good - many thanks for the comments

PR #565, opened 2022-06-13
feat: "skipupdate" model field tag

### LINKED ISSUE #508: How to exclude certain columns when updating
Hi, 

First of all, thank you for the amazing work that you guys are doing! 

I'm currently working on an update query that excludes certain columns, I did find a feature that provides excluding them by using ExcludeColumn. However, I was wondering if there's a feature like "updatable" in JPA, where if you set 'false' to the column that you don't mean to update, it doesn't when updating. Thank you again!

here's the reference for this feature : https://docs.jboss.org/hibernate/jpa/2.1/api/javax/persistence/Column.html

image

funvit:
Closes #508

- [x] Some new tests
- [ ] Documentation?


internal/dbtest/db_test.go
```diff
@@ -1247,6 +1249,48 @@ func testMultiUpdate(t *testing.T, db *bun.DB) {
...
+	model_ := new(Model)
+	model_.ID = model.ID
+	err = db.NewSelect().Model(model_).WherePK().Scan(ctx)
+	require.NoError(t, err, "select")
+	require.NotEmpty(t, model_)
+	require.EqualValues(t, model.ID, model_.ID)
+	require.EqualValues(t, model.Name, model_.Name)
+	require.EqualValues(t, now.UTC(), model_.CreatedAt.UTC())
+
+	require.NotEqualValues(t, model.CreatedAt.UTC(), model_.CreatedAt.UTC())
```
vmihailenco, line 1291: Can you change the test so we save the `CreatedAt` value somewhere and then check that is was not changed:

```go
require.Equal(t, createdAt, model.CreatedAt)
```

It reads a lot easier that way.

query_update.go
```diff
@@ -278,6 +278,9 @@ func (q *UpdateQuery) appendSetStruct(
 		if !hasValue && q.omitZero && f.HasZeroValue(model.strct) {
 			continue
 		}
+		if f.SkipUpdate() {
+			continue
+		}
```
vmihailenco, line 283: Perhaps check this as early as possible?

internal/dbtest/db_test.go
funvit: Done.

query_update.go
funvit: Done.

Issue #291, opened 2021-11-05
Postgres hstore support


dalmarcogd:
Hi guys,
Is there any expectation of giving full support to Postgres hstore and composite? worked great on go-pg https://pg.uptrace.dev/models/

I have been trying to use but in bun he generate hstore column as json...

vmihailenco:
I don't have plans to add support for hstore and composite, but a PR would be welcome.

crashiura:
@vmihailenco Hi, support for hstore only map[string]string ?

PR #567, opened 2022-06-14
Add hstore support

oGi4i:
Adds #291


schema/scan.go
```diff
@@ -449,6 +449,11 @@ func PtrScanner(fn ScannerFunc) ScannerFunc {
 		if dest.IsNil() {
 			dest.Set(reflect.New(dest.Type().Elem()))
 		}
+
+		if dest.Kind() == reflect.Map {
+			return fn(dest, src)
+		}
```
vmihailenco, line 455: This looks confusing and wrong. Do we need it? Is there any other way to achieve the same?

vmihailenco:
@oGi4i thanks - this looks pretty good and clean overall, but I have some concerns about the change in `PtrScanner` function.

schema/scan.go
oGi4i: Well, this covers a pretty odd case, but which I came across in my expirience - usage of `*map[string]string` (there is a test on this in pg_test.go).
Without this it causes a `panic: reflect: call of reflect.Value.Elem on map Value`

vmihailenco: Could you please make sure we have a test for that case so we don't break it accidentally?

oGi4i: Yeah, there is a test on this [here](https://github.com/uptrace/bun/pull/567/files#diff-1475aeec71b3cf89192f35901eb4162d64cc2875904261d1cbcfe60363ed613aR544-R550)

vmihailenco: :+1:

PR #569, opened 2022-06-15
docs: add example for string representation

### LINKED ISSUE #549: String representation of `bun.SelectQuery`
I cannot find on docs and on this Github repo a way to have the query's string representation.

I need to test that many cases (really many because we are exposing columns with possible `OR`, `AND` and `NOT` logical operators) are correctly generating SQL queries.

So I think to add many tests to check if the methods we have generate correct final SQL code.

Is there a way to do this?

If I had a way to generate the final SQL string I could just compare them in tests. End.

I searched the old `go-pg` repository and it seems possible to do this: https://github.com/go-pg/pg/issues/1366#issuecomment-659263194.

sashamelentyev:
Closes #549


example/string-representation/main.go
```diff
@@ -0,0 +1,55 @@
...
+	defer db.Close()
+
+	if err := createSchema(ctx, db); err != nil {
+		panic(err)
+	}
+
+	q := db.NewSelect().Model((*Item)(nil)).Where("id > ?", 0)
+
+	var buf []byte
+	buf, err = q.AppendQuery(schema.NewFormatter(sqlitedialect.New()), nil)
```
vmihailenco, line 36: Please replace `schema.NewFormatter(sqlitedialect.New())` with `db.Formatter()`.

example/string-representation/main.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(string(buf))
+}
+
+func createSchema(ctx context.Context, db *bun.DB) error {
```
vmihailenco, line 44: This can be removed, right?

example/string-representation/main.go
sashamelentyev: Done

example/string-representation/main.go
sashamelentyev: Yes. I removed

vmihailenco:
Thanks! Perhaps as a next step we could also add `q.String()` that uses `q.AppendQuery(db.Formatter(), nil)` internally.

PR #571, opened 2022-06-15
feat: add get where fields

### LINKED ISSUE #534: Access Query Where Fields in Model.BeforeSelect Hook
Is there any capability available to access the query's where fields in a model's BeforeSelect hook? I have a nested struct with a field that I want to make sure is in the where fields of the query before executing and if not return an error. Basically trying to validate the query before execution.

sashamelentyev:
Closes #534

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -6,6 +6,7 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


vmihailenco:
Sorry, but Bun QueryBuilder is write-only. Perhaps you should create your own type instead.

PR #574, opened 2022-06-16
feat: add query string representation

sashamelentyev:


sashamelentyev:
@vmihailenco i did it

query_select.go
```diff
@@ -920,6 +920,17 @@ func (q *SelectQuery) whereExists(ctx context.Context) (bool, error) {
 	return n == 1, nil
 }
 
+func (q *SelectQuery) String() (string, error) {
```
vmihailenco, line 923: Let's just return `string` so we satisfy `Stringer` interface. We can panic on errors.

query_select.go
```diff
@@ -920,6 +920,17 @@ func (q *SelectQuery) whereExists(ctx context.Context) (bool, error) {
...
+	var buf []byte
```
vmihailenco, line 924: Nitpick: no need to define `buf`.

sashamelentyev:
@vmihailenco done

vmihailenco:
Thank you :+1:

PR #575, opened 2022-06-16
feat: add get where fields

sashamelentyev:
Closes: #534 

@vmihailenco mb this?


vmihailenco:
@sashamelentyev Yes, example would not hurt.

Issue #258, opened 2021-10-19
does bun support mysql `USE INDEX` except raw sql


shawn246:
https://dev.mysql.com/doc/refman/5.7/en/index-hints.html

PR #577, opened 2022-06-16
Issue 258: Add index hints for select and update

VladimirStepanov:
Close https://github.com/uptrace/bun/issues/258


VladimirStepanov:
@vmihailenco Hello! I added some functionality, could you check it plz? And I have a question:
Index hints is supporting only for MySQL. What I should do for other databases?

vmihailenco:
:+1: 

>What I should do for other databases?

Just ignore the hints when they are not supported.

query_select.go
```diff
@@ -159,6 +160,13 @@ func (q *SelectQuery) WhereAllWithDeleted() *SelectQuery {
 
 //------------------------------------------------------------------------------
 
+func (q *SelectQuery) IndexHints(indexes ...string) *SelectQuery {
```
vmihailenco, line 163: Let's rename this to `UseIndex`, because there is also `ForceIndex` and `IgnoreIndex`

query_base.go
```diff
@@ -1096,3 +1096,40 @@ func (q cascadeQuery) appendCascade(fmter schema.Formatter, b []byte) []byte {
 	}
 	return b
 }
+
+//------------------------------------------------------------------------------
+
+type idxHintsQuery struct {
+	idxHints [][]schema.QueryWithArgs
```
vmihailenco, line 1103: Can't we flatten this `[]schema.QueryWithArgs`? Not sure if we need to support multiple `USE INDEX` clauses...

vmihailenco:
Please also add some tests to https://github.com/uptrace/bun/blob/master/internal/dbtest/query_test.go

query_base.go
VladimirStepanov: Yeah, I agree
But here is https://dev.mysql.com/doc/refman/5.7/en/index-hints.html has interesting example:

```sql
SELECT * FROM t1 USE INDEX (i1) USE INDEX (i1,i1);
```

VladimirStepanov:
@vmihailenco I want to add ForceIndex and IgnoreIndex, is it ok if I do it in this PR?

query_base.go
vmihailenco: The doc says that it is not an error, but it also does not say that it is useful... is it useful?

internal/dbtest/testdata/snapshots/TestQuery-mariadb-131
```diff
@@ -0,0 +1 @@
+SELECT `id` FROM `models` AS `model` USE INDEX (`ix1, ix2`)
```
vmihailenco, line 1: This looks wrong - I don't think you want to quote 2 index names together ``` `ix1, ix2` ```

query_base.go
```diff
@@ -1096,3 +1096,137 @@ func (q cascadeQuery) appendCascade(fmter schema.Formatter, b []byte) []byte {
...
+	uIndex           [][]schema.QueryWithArgs
+	uIndexForJoin    [][]schema.QueryWithArgs
+	uIndexForOrderBy [][]schema.QueryWithArgs
+	uIndexForGroupBy [][]schema.QueryWithArgs
+
+	// ignore index
+	ignIndex           [][]schema.QueryWithArgs
+	ignIndexForJoin    [][]schema.QueryWithArgs
+	ignIndexForOrderBy [][]schema.QueryWithArgs
+	ignIndexForGroupBy [][]schema.QueryWithArgs
```
vmihailenco, line 1113: Let's not shorten variables like this (`u` => `use`) - it is uncommon and hard to understand.

Perhaps this should be:

```
type SelectQuery struct {
    indexHintsQuery
}

type indexHintsQuery struct {
    // lazily allocated when index is added
    use     *indexHints  
    ignore *indexHints
}

type indexHints struct {
	names         []schema.QueryWithArgs
	forJoin         []schema.QueryWithArgs
	forOrderBy  []schema.QueryWithArgs
	forGroupBy []schema.QueryWithArgs
}
```

What do you think?

query_base.go
VladimirStepanov: Yeah, you are right

internal/dbtest/testdata/snapshots/TestQuery-mariadb-131
VladimirStepanov: Heh, passed into tests a string "`ix1, ix2`" instead "(ix1", "ix2")
:)

query_base.go
VladimirStepanov: Yes, good idea
Fixed

VladimirStepanov:
@vmihailenco Fuuh, I added support for all index hints (use, ignore, force)

query_base.go
```diff
@@ -1096,3 +1096,252 @@ func (q cascadeQuery) appendCascade(fmter schema.Formatter, b []byte) []byte {
...
+
+func (ih *idxHintsQuery) appendUseIndex(indexes ...string) {
+	if len(indexes) == 0 {
+		return
+	}
+	if ih.use == nil {
+		ih.use = new(indexHints)
+	}
+	ih.use.names = ih.addIdxHints(indexes...)
+}
```
vmihailenco, line 1131: The func name `appendUseIndex` says append, but instead it overwrites `names`. I think this can be simplified like this:

```go
func (ih *idxHintsQuery) appendUseIndex(indexes ...string) {
    use := ih.lazyUse()
    use.names = appendIndexHints(use.names, indexes...)
}

func (ih *idxHintsQuery) lazyUse() *indexHints {
	if ih.use == nil {
		ih.use = new(indexHints)
	}
	return ih.use
}

func appendIndexHints(hints []schema.QueryWithArgs, indexes ...string) []schema.QueryWithArgs {
	for _, idx := range indexes {
		hints = append(hints, schema.UnsafeIdent(idx))
	}
	return hints
}
```

What do you think?

vmihailenco:
Is there a specific reason why `UseIndex` discard previous indexes? For example, the following behavior is surprising and inconsistent so we need a good reason/explanation why it is done so:

```go
q.Table("one").Table("two") // generates FROM one, two

q.UseIndex("one").UseIndex("two") // generates USE INDEX (two)
```

VladimirStepanov:
@vmihailenco Yeah, it was a bug. Fixed

query_base.go
```diff
@@ -1096,3 +1096,235 @@ func (q cascadeQuery) appendCascade(fmter schema.Formatter, b []byte) []byte {
...
+			},
+			{
+				Name:   "USE INDEX FOR ORDER BY",
+				Values: ih.use.forOrderBy,
+			},
+			{
+				Name:   "USE INDEX FOR GROUP BY",
+				Values: ih.use.forGroupBy,
+			},
+		}...)
```
vmihailenco, line 1254: I don't want to test your patience, but you could write this in a shorter and faster way:

```go
if ih.use != nil {
    b = appendIndexHints(fmter, b, "USE INDEX", ih.use)
}
if ih.ignore != nil {
    b = appendIndexHints(fmter, b, "IGNORE INDEX", ih.ignore)
}
if ih.ignore != nil {
    b = appendIndexHints(fmter, b, "FORCE", ih.force)
}
```

Then in `appendIndexHints`:

```go
func appendIndexHints(fmter Formatter, b []byte, prefix string, ih *indexHints) {
    b = appendIndexHint(fmter, b, prefix, "", ih.names)
    b = appendIndexHint(fmter, b, prefix, "FOR JOIN", ih.forJoin)
    b = appendIndexHint(fmter, b, prefix, "FOR ORDER BY", ih.forOrderBy)
    b = appendIndexHint(fmter, b, prefix, "FOR GROUP BY", ih.forGroupBy)
    return b
}
```

Would be nice you could send a cleanup PR :+1:

query_base.go
```diff
@@ -1096,3 +1096,235 @@ func (q cascadeQuery) appendCascade(fmter schema.Formatter, b []byte) []byte {
...
+func (ih *idxHintsQuery) bufIndexHint(
+	name string,
+	hints []schema.QueryWithArgs,
+	fmter schema.Formatter, b []byte,
+) ([]byte, error) {
+	var err error
+	if len(hints) == 0 {
+		return b, nil
+	}
+	b = append(b, fmt.Sprintf(" %s (", name)...)
```
vmihailenco, line 1318: You could save an allocation by avoiding `fmt.Sprintf` and using several `append` calls instead.

vmihailenco:
Thanks for your work :+1: This is good enough so I am merging this, but please consider https://github.com/uptrace/bun/pull/577#discussion_r909249720

query_base.go
VladimirStepanov: Thank you! 
I will send changes in another PR

PR #581, opened 2022-06-17
docs: add get where fields example

sashamelentyev:
Closes: #534

@vmihailenco example for case when we need get where data


sashamelentyev:
@vmihailenco check this please

vmihailenco:
Thanks

Issue #588, opened 2022-06-21
MSSQL can't use limit when offset is 0


lstrihic:
Hi,

I can't use offset https://github.com/uptrace/bun/blob/e5d8fe805e6a2a78feaa9221539768cbe40247e9/query_select.go#L508 because of the this statement.

PR #589, opened 2022-06-21
fix: mssql cursor pagination

lstrihic:
#588

Here is the fix for MSSQL when offset is set to `0`.

### DIFF
--- a/internal/dbtest/orm_test.go
+++ b/internal/dbtest/orm_test.go
@@ -127,6 +127,7 @@ func testAuthorRelations(t *testing.T, db *bun.DB) {
 		Relation("Books.Translations", func(q *bun.SelectQuery) *bun.SelectQuery {
 			return q.OrderExpr("tr.id ASC")
 		}).
+		OrderExpr("author.id ASC").
 		Limit(1).
 		Scan(ctx)
 	require.NoError(t, err)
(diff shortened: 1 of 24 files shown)


vmihailenco:
Thanks :+1: Could you add a test to https://github.com/uptrace/bun/blob/master/internal/dbtest/query_test.go#L789-L800?

lstrihic:
I just run it locally and everything seems to be fine.
image

lstrihic:
> > The test failure looks real:
> > ```
> > --- FAIL: TestORM (2.64s)
> >     --- FAIL: TestORM/mssql2019 (0.34s)
> >         --- FAIL: TestORM/mssql2019/testAuthorRelations (0.00s)
> >             orm_test.go:132: 
> >                 	Error Trace:	orm_test.go:132
> >                 	            				orm_test.go:43
> >                 	Error:      	Received unexpected error:
> >                 	            	mssql: Invalid usage of the option NEXT in the FETCH statement.
> >                 	Test:       	TestORM/mssql[201](https://github.com/uptrace/bun/runs/6998766509?check_suite_focus=true#step:5:202)9/testAuthorRelations
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any ideas what is wrong?
> 
> I have to check.

In mysql to use limit we have to have defined order.

vmihailenco:
Thanks :+1:

Issue #587, opened 2022-06-20
Unit testing service logic when using - *DB.RunInTx func


GuyKomari:
I have a service with `*bun.DB` instance and a method that handles a transaction using `RunInTx` func.
I would like to add a unit test for this method but in order to do that, I need to use an abstraction over `*bun.DB` 
which seems to be - `bun.IDB` interface.
However,  `bun.IDB` is missing `RunInTx` func and other "transaction-related" methods like - BeginTx, Begin, etc.

simple code example:
```
type MyService struct {
	db            *bun.DB
	myRepo        MyRepo
}

func (s *MyService) DoSmt(ctx context.Context) error {
	err := s.db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
	        // Get for update
                // Some logic
                // Update
               
               return nil
	})
	
        return err
}

type MyRepo interface { 
     Get(ctx context.Context, db bun.IDB) interface{}, error
     Update(ctx context.Context, db bun.IDB, interface{}) interface{}, error
}
```

Currently, to overcome this issue, I have created this wrapper that can be used instead of `*bun.DB`:
```
type DbWrapper interface {
	bun.IDB
	RunInTx(ctx context.Context, opts *sql.TxOptions, fn func(ctx context.Context, tx bun.Tx) error) error
}
```
However, it would be nice if it was part of `bun.DBI`.
Would it be possible to add the "transaction-related" methods to `bun.DBI` (same as in go-pg)? 
Is there a better approach to handle this case?

TIA, Guy

vmihailenco:
IDB is a common interface for *bun.DB, bun.Conn, and bun.Tx. I don't think we can extend it with `Begin` and `RunInTx`.

But we probably can add another interface, for example, `ITx` in addition to `IConn`. Then we can probably create `ITxDB` in addition to `IDB`. Not sure if it is worth it and makes your task much easier...

isgj:
Some other ORMs offer the same interface also for the transaction, and when you are already in a tx they use `SAVEPOINT`.

The `bun.Tx` can be extended with `Begin`, `RunInTx`.. which will work on save-points. In this case the interface `IDB` can be extended and other than solving this issue it will make it possible for `bun.DB` and `bun.Tx` to be used interchangeably (helpful in some scenarios).

vmihailenco:
@isgj yes, that is an option too. If there is any interest, please send a PR.

PR #591, opened 2022-06-26
feat: add tx methods to IDB (#587)

isgj:
Updated `IDB` interface with 2 methods

```
BeginTx(ctx context.Context, opts *sql.TxOptions) (Tx, error)
RunInTx(ctx context.Context, opts *sql.TxOptions, f func(ctx context.Context, tx Tx) error) error
```

When creating a savepoint the `opts` are ignored.
Currently the transaction will not use the name but we can store something like `TX_{random}` same format as for savepoint. Then we can also emit the name with the hooks so logging libraries can show if a query is being executed as part of a transaction or savepoint.

### DIFF
--- a/db.go
+++ b/db.go
@@ -2,7 +2,9 @@ package bun
 
 import (
 	"context"
+	"crypto/rand"
 	"database/sql"
+	"encoding/hex"
 	"fmt"
 	"reflect"
 	"strings"
@@ -431,6 +433,8 @@ func (db *DB) PrepareContext(ctx context.Context, query string) (Stmt, error) {
 type Tx struct {
 	ctx context.Context
 	db  *DB
+	// name is the name of a savepoint
+	name string
 	*sql.Tx
 }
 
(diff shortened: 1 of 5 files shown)


db.go
```diff
@@ -534,6 +564,54 @@ func (tx Tx) QueryRowContext(ctx context.Context, query string, args ...interfac
 
 //------------------------------------------------------------------------------
 
+func (tx Tx) Begin() (Tx, error) {
+	return tx.BeginTx(context.Background(), nil)
+}
+
+// BeginTx will save a point in the running transaction.
+func (tx Tx) BeginTx(ctx context.Context, _ *sql.TxOptions) (Tx, error) {
+	uid, err := uuid.NewRandom()
```
vmihailenco, line 573: Let's not add a dependency on `uuid` just to generate a random string.

```go
    b := make([]byte, 16)
    rand.Read(b)
```

vmihailenco:
This looks pretty good. Could you add a test for savepoints in https://github.com/uptrace/bun/blob/master/internal/dbtest/db_test.go#L273?

db.go
isgj: I used `crypto/rand` but I think also `math/rand` is an option as it is safe for concurrent usage (for this usage pseudo random is fine). Maybe we can check which is faster but that will be one word change.

isgj:
`mssql` has a different syntax https://docs.microsoft.com/en-us/sql/t-sql/language-elements/save-transaction-transact-sql?view=sql-server-ver16

How are these cases handled when different sql should be generated?

vmihailenco:
>How are these cases handled when different sql should be generated?

You can add a feature flag for mssql [here](https://github.com/uptrace/bun/blob/master/dialect/mssqldialect/dialect.go#L41-L51) and use it like [this](https://github.com/uptrace/bun/blob/master/query_update.go#L199-L205)

vmihailenco:
Excellent, thank you :+1:

Issue #593, opened 2022-06-27
Add ability to run and scan raw queries


vmihailenco:
Something like this

```go
err := db.Raw(query, args...).Scan(ctx, &dest)
```

See https://github.com/uptrace/bun/discussions/592 for context

nanmu42:
This is a rather cool feature. :+1:  I wonder, when will it be released?

vmihailenco:
Just tagged a new version

PR #596, opened 2022-06-28
feat: added raw query calls

JoseARomeroJ:
Added raw query support: https://github.com/uptrace/bun/issues/593

### DIFF
--- a/query_raw.go
+++ b/query_raw.go
@@ -39,10 +39,7 @@ func (q *RawQuery) Scan(ctx context.Context, dest ...interface{}) error {
 }
 
 func (q *RawQuery) AppendQuery(fmter schema.Formatter, b []byte) ([]byte, error) {
-	query := []byte(fmter.FormatQuery(q.query, q.args))
-	b = append(b, query...)
-
-	return b, nil
+	return fmter.AppendQuery(b, q.query, q.args), nil
 }
 
 func (q *RawQuery) Operation() string {
--- a/query_raw.go
+++ b/query_raw.go
@@ -1,0 +1,50 @@
+package bun
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


vmihailenco:
Hey,

The suggested API looks like this:

```go
err := db.Raw(query, args...).Scan(ctx, &strct)
err := db.Raw(query, args...).Scan(ctx, &slice)
err := db.Raw(query, args...).Scan(ctx, &var1, &var2)
```

The implementation should look like this:

```go
func (db *DB) Raw(query string, args ...interface{}) *RawQuery {
    return &RawQuery{
        db: db,
        query: query,
        args: args,
    }
}

type RawQuery struct {
    db *DB
    query string
    args ...interface{}
}

func (q *RawQuery) Scan(ctx context.Context, dest ...interface{}) error {
    // extract and reuse Scan implementation from SelectQuery.Scan
    // to format the query, use query := db.format(q.query, q.args)
    return scanModel(...)
}
```

Does that make sense?

JoseARomeroJ:
Did you mean something like that?

query_raw.go
```diff
@@ -0,0 +1,50 @@
...
+	query := q.db.format(q.query, q.args)
+	_, err = q.scan(ctx, q, query, model, true)
+	return err
+}
+
+func (q *RawQuery) AppendQuery(fmter schema.Formatter, b []byte) ([]byte, error) {
+	query := []byte(fmter.FormatQuery(q.query, q.args))
+	b = append(b, query...)
+
+	return b, nil
```
vmihailenco, line 45: This can be `return fmter.AppendQuery(b, q.query, q.args), nil`

vmihailenco:
Very good - thanks :+1:

PR #599, opened 2022-07-04
feat: create an extra module for newrelic

l-farrell:
Similar to the OTEL module which exists.


vmihailenco:
What if we rename the module to `bunrelic` and then rename this method to `NewQueryHook` so the API is similar:

```go
bunotel.NewQueryHook(...)

->

bunrelic.NewQueryHook(...)
```

Also, please  add a compile-time interface [assertion](https://medium.com/@matryer/golang-tip-compile-time-checks-to-ensure-your-type-satisfies-an-interface-c167afed3aae):

```go
var _ bun.QueryHook = (*QueryHook)(nil)
```

vmihailenco:
Thanks :+1:

PR #602, opened 2022-07-04
feat(pgdialect): add identity support

### LINKED ISSUE #479: struct tags "autoincrement"
## PostgreSQL

[Don't use serial](https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_serial)

> For new applications, **identity columns** should be used instead.

oGi4i:
Closes #479


oGi4i:
@vmihailenco thanks for your thoughts on this, please check out this implementation

query_table_create.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/uptrace/bun/schema"
 )
 
+type IdentityAppender interface {
+	AppendIdentity(b []byte) []byte
+}
```
vmihailenco, line 17: Seems to be unused. Let's remove.

oGi4i: Yeah, sorry, forgot to remove that. Done.

oGi4i:
> @oGi4i looks good 👍 Could you also add a test to https://github.com/uptrace/bun/blob/master/internal/dbtest/query_test.go#L899 so we are sure the generated query stays the same?

Thanks.
Added a test case to check that create table query is generated correctly for all dialects

vmihailenco:
Thanks :+1:

PR #608, opened 2022-07-09
fix: bunotel add set attributes to query metrics

guyguy333:
Currently, even if we set attributes with `WithAttributes`, they are not set for query metrics, only pool/connections metrics. This PR fixes this issue.


vmihailenco:
Thanks

PR #632, opened 2022-07-29
Fix/composite in

### LINKED ISSUE #626: Has-Many relation with composite keys throws error on mssql
mssql does not like `IN` conditions with multiple fields and that messes things up when using `has-many` relations with more than 1 join.
In the example below ` (("contact"."company_no", "contact"."department_no") IN (('ABC', 'AA')))` generates the error.
`mssql: An expression of non-boolean type specified in a context where a condition is expected, near ','`

See https://stackoverflow.com/questions/35194028/in-clause-in-sql-server-with-multiple-columns or https://stackoverflow.com/a/1136381/232715 for a possible alternative ways of doing it for mssql.

```
[bun]  13:53:47.583   SELECT                1.617ms  SELECT "contact"."company_no", "contact"."department_no", "contact"."name" FROM "contacts" AS "contact" WHERE (("contact"."company_no", "contact"."department_no") IN (('ABC', 'AA')))          mssql.Error: mssql: An expression of non-boolean type specified in a context where a condition is expected, near ','. 
```

Relevant models
```
type Department struct {
	CompanyNo string `bun:"company_no,pk"`
	No        string `bun:"no,pk"`
	Name      string
	Contacts  []*Contact `bun:"rel:has-many,join:company_no=company_no,join:no=department_no"`
}

type Contact struct {
	bun.BaseModel
	CompanyNo    string `bun:"company_no,pk"`
	DepartmentNo string `bun:"department_no,pk"`
	Name         string `bun:"name,pk"`
}
```

Full example source in gist with both sqlite and mssql (that throws error) in the same file.
https://gist.github.com/kmpm/22450dfd53a2adeb11a2deccca1e7d06

kmpm:
Closes #626 for me and doesn't fail any other tests.

### DIFF
--- a/relation_join.go
+++ b/relation_join.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"reflect"
 
+	"github.com/uptrace/bun/dialect/feature"
 	"github.com/uptrace/bun/internal"
 	"github.com/uptrace/bun/schema"
 )
@@ -60,6 +61,14 @@ func (j *relationJoin) manyQuery(q *SelectQuery) *SelectQuery {
 	q = q.Model(hasManyModel)
 
 	var where []byte
+
+	if q.db.dialect.Features().Has(feature.CompositeIn) {
+		return j.manyQueryCompositeIn(where, q)
+	}
+	return j.manyQueryMulti(where, q)
+}
+
+func (j *relationJoin) manyQueryCompositeIn(where []byte, q *SelectQuery) *SelectQuery {
 	if len(j.Relation.JoinFields) > 1 {
 		where = append(where, '(')
 	}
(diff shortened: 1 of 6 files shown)


relation_join.go
```diff
@@ -312,3 +344,64 @@ func appendChildValues(
...
+	return list
+}
+
+func appendMultiValues(
+	fmter schema.Formatter, b []byte, v reflect.Value, index []int, baseFields, joinFields []*schema.Field, table schema.Safe,
+) []byte {
+	// This is a mix of appendChildValues and query_base.appendColumns
+	if len(joinFields) != len(baseFields) {
+		panic("asdfasdf")
+	}
```
vmihailenco, line 370: Let's remove or use `panic("not reached")`

relation_join.go
```diff
@@ -312,3 +344,64 @@ func appendChildValues(
...
+			b = append(b, '.')
+		}
+		b = append(b, f.SQLName...)
+		list = append(list, b)
+	}
+	return list
+}
+
+func appendMultiValues(
+	fmter schema.Formatter, b []byte, v reflect.Value, index []int, baseFields, joinFields []*schema.Field, table schema.Safe,
```
vmihailenco, line 365: Let's rename `table` to `joinTable` for clarity

relation_join.go
```diff
@@ -312,3 +344,64 @@ func appendChildValues(
...
+	walk(v, index, func(v reflect.Value) {
+		start := len(b)
+		for i, f := range baseFields {
+			if i > 0 {
+				b = append(b, " AND "...)
+			}
+			if len(baseFields) > 1 {
+				b = append(b, '(')
+			}
+			b = append(b, joins[i]...)
```
vmihailenco, line 386: Looks like we can get rid of `getColumns` and instead write here:

```go
		b = append(b, joinTable...)
		b = append(b, '.')
		b = append(b, joinFields[i].SQLName...)
```

What do you think?

vmihailenco:
I've left some minor comments, but overall this looks very good :+1:

kmpm:
Sorry for taking so long to resolve your comments which all made good sense.

vmihailenco:
Thanks you :+1:

Issue #525, opened 2022-04-26
Getting randomly EOF error


xhkocatepe:
We receive sometimes an EOF error from bun as ` *errors.errorString: EOF` .
It is important that we have get issue `sometimes` ; same query, same db, same data, sometimes return data, but sometimes get `EOF` error. 
We get a workaround solution for using retry policy. We try to run same query until we successfully get data from db.

**Is there any real solution for this issue?**

PS:If it is necessary, I can share part of query.

xhkocatepe:
Hello Everyone :) Is there any response?

hhuseyinpay:
Which driver do you use? If you use the pgdriver, there is a hint in [the documentation ](https://bun.uptrace.dev/postgres/#debugging)

xhkocatepe:
yeah we are using pgdriver 1.0.19. It is not working shared file from you.

vmihailenco:
Closing, because there is no way to reproduce the problem. Please re-open if you have more details.

xhkocatepe:
We can reproduce the problem many times. What kind of details dou you want?

htdvisser:
I've been able to reproduce this issue: https://gist.github.com/htdvisser/fe1145e832fe2f323ee609e866fb59be

I'm investigating further and will try to contribute a fix.

hypnoglow:
We still getting a lot of random EOFs with the 1.1.9 bun version. I have no reproducer though at this point.

PR #633, opened 2022-07-30
fix(pgdriver): return first error if readMessageType errors

htdvisser:
This pull request contributes a fix for https://github.com/uptrace/bun/issues/525 that I was able to reproduce by making PostgreSQL close client connections (first revision of https://gist.github.com/htdvisser/fe1145e832fe2f323ee609e866fb59be).

Before the server closes the client connection, it sends a `57P01 terminating connection due to administrator command`. This error is however not returned by the read loops, which instead return the EOF coming from trying to read the next message.

~This pull request changes the read loops to check if there's a `firstErr` to return before returning the error returned by `readMessageType` (which in my experiment was usually a `read tcp [::1]:***->[::1]:5432: read: connection reset by peer`, but occasionally an `EOF`).~

EDIT: Updated, see comments.

I've verified that the seemingly random EOFs no longer occur in the last revision of https://gist.github.com/htdvisser/fe1145e832fe2f323ee609e866fb59be that replaces `pgdriver` with my fork.

### DIFF
--- a/driver/pgdriver/proto.go
+++ b/driver/pgdriver/proto.go
@@ -1026,7 +1026,14 @@ func readError(rd *reader) (error, error) {
 		}
 		m[c] = s
 	}
-	return Error{m: m}, nil
+	switch err := (Error{m: m}); err.Field('V') {
+	case "FATAL", "PANIC":
+		// Return this as an error and stop processing.
+		return nil, err
+	default:
+		// Return this as an error message and continue processing.
+		return err, nil
+	}
 }
 
 //------------------------------------------------------------------------------


vmihailenco:
Hi @htdvisser,

Thanks for digging into this.

It looks like pgx just stops reading closes the connection on FATAL errors - see https://github.com/jackc/pgconn/blob/fe0fb3b24dca7c395038dc81bdb84e6706e50979/pgconn.go#L568-L573 . I think we should try to do the same in pgdriver.

What about if we change readError like this:

```go
func readError(cn *conn, rd *reader) (error, error) {
	m := make(map[byte]string)

	for {
		c, err := rd.ReadByte()
		if err != nil {
			return nil, err
		}
		if c == 0 {
			break
		}
		s, err := readString(rd)
		if err != nil {
			return nil, err
		}
		m[c] = s
	}

	switch err := Error{m: m}; err.Field('V') {
	case "FATAL", "PANIC":
		 // Return this as an error to stop processing.
		 return nil, err
	default:
		// Return this as an error message and continue processing.
		return err, nil
	}
}
```

What do you think?

htdvisser:
I'm not familiar with PostgreSQL internals, but I think we can indeed assume that FATAL errors always mean the end of the connection, so it makes sense to just stop processing and mark the connection as "bad".

Making pgdriver treat these errors the same as pgx does, will also result in less "surprises" for users, so I think your solution may be better.

I've committed your proposed solution (removing the unused `cn *conn`) in my fork (https://github.com/uptrace/bun/compare/master...htdvisser:forked-bun:633-eof-alt) and re-ran my tests (third revision of https://gist.github.com/htdvisser/fe1145e832fe2f323ee609e866fb59be); the errors are now indeed as expected.

I suspect you wanted to use that `cn *conn` to mark the connection as "bad"?

vmihailenco:
Looks good - thanks!

>I suspect you wanted to use that cn *conn to mark the connection as "bad"?

Exactly, but looks like we already handle that on a upper level so no need.

htdvisser:
Good. Thank you!

PR #640, opened 2022-08-06
feat: add raw query to IDB

choyri:
Optimize #596


zekth:
@vmihailenco could this be released?

PR #679, opened 2022-09-22
fix(query): make WhereDeleted compatible with ForceDelete

### LINKED ISSUE #673: (Accidental?) Change in `ForceDelete()` behavior vs. go-pg/pg
Given a `Model` using the soft deletion feature:

```go
type Model struct {
	// [...]
	DeletedAt *time.Time `bun:",soft_delete"` // previously `pg:",soft_delete"`
}
```

#### Previously in [go-pg/pg](https://github.com/go-pg/pg)

The query `db.Model((*Model)(nil)).ForceDelete()` generated `DELETE FROM model WHERE deleted_at IS NOT NULL`, i.e. performed the **hard deletion of all soft deleted rows**.

This is because the `deletedFlag` was [forcefully set](https://github.com/go-pg/pg/blob/v10.10.6/orm/query.go#L1207) and taken into account when generating the WHERE clause ([isSoftDelete returns true](https://github.com/go-pg/pg/blob/v10.10.6/orm/query.go#L236), [appendSoftDelete is called](https://github.com/go-pg/pg/blob/v10.10.6/orm/query.go#L1439), and the [clause is added](https://github.com/go-pg/pg/blob/v10.10.6/orm/query.go#L1449)). 

#### Now with [bun](https://github.com/uptrace/bun)

The query `db.NewDelete().Model((*Model)(nil)).ForceDelete().Exec(ctx)` generates `DELETE FROM model`, i.e. performs the **hard deletion of all rows, whether soft deleted or not**.

Indeed:
1. In `(q *DeleteQuery) AppendQuery`, the `deletedFlag` is still forcefully applied to the query ([query_delete.go#L161](https://github.com/uptrace/bun/blob/v1.1.8/query_delete.go#L161)), but
2. The `deletedFlag` is then ignored by `(q *whereBaseQuery) appendWhere` ([query_base.go#L778](https://github.com/uptrace/bun/blob/v1.1.8/query_base.go#L778)) because the new implementation of `(q *DeleteQuery) isSoftDelete()` returns `false` ([query_delete.go#L205](https://github.com/uptrace/bun/blob/v1.1.8/query_delete.go#L205) due to `forceDeleteFlag` being set).

I wonder if that change was intentional or accidental, as forcing a flag to be set that is not then used seems to be a mistake.

It is possible to restore previous behavior by manually adding a `Where("deleted_at IS NOT NULL")` clause to the query, but using `WhereDeleted()` explicitly is not an option since the `deletedFlag` would still be ignored.

IMHO, keeping the new behavior but letting `WhereDeleted()` work in conjunction with `ForceDelete()` would be the nicest option here:

```go
db.NewDelete().Model((*Model)(nil)).ForceDelete().Exec(ctx)
// Generates: DELETE FROM model

db.NewDelete().Model((*Model)(nil)).WhereDeleted().ForceDelete().Exec(ctx)
// Generates: DELETE FROM model
// I would like instead: DELETE FROM model WHERE deleted_at IS NOT NULL
```

What do you think?

maximerety:
As of Bun 1.1.8, `WhereDeleted()` is ignored when used with `ForceDelete()`.

This commit fixes that problem:

```go
db.NewDelete().Model((*Model)(nil)).Where("1 = 1").WhereDeleted().ForceDelete().Exec(ctx)
// Before this commit: DELETE FROM model WHERE 1 = 1
// After this commit:  DELETE FROM model WHERE 1 = 1 AND deleted_at IS NOT NULL
```

Additionally, the `Where("1 = 1")` trick is no longer required as `WhereDeleted()` now counts as a user-defined `WHERE` clause in `mustAppendWhere`:

```go
db.NewDelete().Model((*Model)(nil)).WhereDeleted().ForceDelete().Exec(ctx)
// Before this commit: Errors with "bun: Update and Delete queries require at least one Where"
// After this commit:  DELETE FROM model WHERE deleted_at IS NOT NULL
```

This gives us an easy and natural way to hard delete all soft deleted records.

Fixes #673.

### DIFF
--- a/internal/dbtest/soft_delete_test.go
+++ b/internal/dbtest/soft_delete_test.go
@@ -111,15 +111,48 @@ func testSoftDeleteForce(t *testing.T, db *bun.DB) {
 	err := db.ResetModel(ctx, (*Video)(nil))
 	require.NoError(t, err)
 
-	video1 := &Video{
-		ID: 1,
+	videos := []Video{
+		{Name: "video1"},
+		{Name: "video2"},
+		{Name: "video3"},
 	}
-	_, err = db.NewInsert().Model(video1).Exec(ctx)
+
+	_, err = db.NewInsert().Model(&videos).Exec(ctx)
 	require.NoError(t, err)
 
-	_, err = db.NewDelete().Model(video1).WherePK().ForceDelete().Exec(ctx)
+	// Force delete video1.
+	_, err = db.NewDelete().Model((*Video)(nil)).Where("name = ?", "video1").ForceDelete().Exec(ctx)
 	require.NoError(t, err)
 
+	// Soft delete video2.
+	_, err = db.NewDelete().Model((*Video)(nil)).Where("name = ?", "video2").Exec(ctx)
+	require.NoError(t, err)
+
+	// Check one visible video.
+	var res []Video
+	err = db.NewSelect().Model((*Video)(nil)).Column("name").Scan(ctx, &res)
+	require.NoError(t, err)
+	require.Equal(t, []Video{{Name: "video3"}}, res)
+
+	// Check one soft deleted video.
+	err = db.NewSelect().Model((*Video)(nil)).Column("name").WhereDeleted().Scan(ctx, &res)
+	require.NoError(t, err)
+	require.Equal(t, []Video{{Name: "video2"}}, res)
+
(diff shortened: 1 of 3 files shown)


vmihailenco:
Thanks :+1: The code looks good, but I was worried about backwards compatibility and that's why the delay. Let's try this out.

PR #693, opened 2022-10-19
feat: Add WithRecursive operation to query

### LINKED ISSUE #526: Support for recusive CTE
Hello, first of all thanks a lot for all the work you've put in go-pg and bun over the years ;)  

I'm trying to use create a query with a recursive CTE. I couldn't find any information about how to do such a thing with Bun so my naive approach was to write something like this :  
```go
err := Db.NewSelect().
            ModelTableExpr("sub_graph").
            With("RECURSIVE sub_graph", subQuery).
            Scan(ctx, &idSlice)
```
Unfortunately the whole string passed to the With() method ends up being quoted in the generated query (which is to be expected considering that the passed string should be a table name).  
I would really appreciate if you could consider slightly changing the way With() works to allow for the generation of recursive CTEs (or adding a WithRecursive() method, it would work just as well for me).

Have a good day !

Nathavee-mh:
close #526 

implement method `WithRecursive` for delete insert select and update query. 
```diff
+ func (q *Query) WithRecursive(name string, query schema.QueryAppender) *Query
```

for recursive query
```SQL
WITH RECURSIVE ... (
...
)
```

Ex.
```SQL
WITH RECURSIVE subordinate AS (
    SELECT
        c.id
    FROM
        categories c
    WHERE
        c.id = '7d82f83b-c45e-4020-8f78-dbfda314e684'
    UNION
    all
    SELECT
        c2.id
    FROM
        categories c2
        INNER JOIN subordinate s ON s.id = c2.parent_category_id
)
SELECT
    DISTINCT subordinate.*
FROM
    subordinate
```
```go
subordinate := db.NewSelect().Table("categories").
	Column("categories.id").
	Where("categories.id = ?", id).
	UnionAll(
		db.NewSelect().Table("categories").
			Column("categories.id").
			Join("INNER JOIN subordinate ON subordinate.id = categories.parent_category_id"),
	)
db.NewSelect().
WithRecursive("subordinate",subordinate).
Table("categories").
Join("LEFT JOIN subordinate on true").
Where("categories.id = subordinate.id")
```

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -20,8 +20,9 @@ const (
 )
 
 type withQuery struct {
-	name  string
-	query schema.QueryAppender
+	name      string
+	query     schema.QueryAppender
+	recursive bool
 }
 
 // IConn is a common interface for *sql.DB, *sql.Conn, and *sql.Tx.
@@ -244,10 +245,11 @@ func (q *baseQuery) isSoftDelete() bool {
 
 //------------------------------------------------------------------------------
 
-func (q *baseQuery) addWith(name string, query schema.QueryAppender) {
+func (q *baseQuery) addWith(name string, query schema.QueryAppender, recursive bool) {
 	q.with = append(q.with, withQuery{
-		name:  name,
-		query: query,
+		name:      name,
+		query:     query,
+		recursive: recursive,
 	})
 }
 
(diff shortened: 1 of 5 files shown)


vmihailenco:
Looks good - thanks!

PR #718, opened 2022-11-23
feat:  mssql output support for update or delete query

### LINKED ISSUE #717: Add feature.Ouput support for update or delete query
It is a fancy feature for MSSQL which could scan the output after update or delete query

bgdnxt:
close #717


vmihailenco:
Looks good, thanks :+1:

PR #719, opened 2022-11-24
Allow QueryEvent to better detect operations in raw queries

maximerety:
When writing long raw queries, it is common to use multiline strings that can start with various white space characters (`\n`, `\t`, space...). E.g.

```go
err := handle.NewRaw(`
	SELECT
		...
	FROM
		...
	LEFT JOIN LATERAL (
		...
	) AS sub ON true
	WHERE
		...
`).Scan(ctx)
```

In that case, a `QueryEvent` would report its `Operation()` as `"\n\tSELECT"`.

This commit gives the ability to report `"SELECT"` instead.

Side-note: some other common cases are not handled anyway (e.g., raw queries with CTE), but this fix would still reduce the number of unexpected results.

### DIFF
--- a/hook.go
+++ b/hook.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"sync/atomic"
 	"time"
+	"unicode"
 
 	"github.com/uptrace/bun/schema"
 )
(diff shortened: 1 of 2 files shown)


vmihailenco:
Thanks

Issue #456, opened 2022-02-22
Add something like GORM's AutoMigrate


vmihailenco:
First, we should provide some unified way to inspect current database schema via using `INFORMATION_SCHEMA.TABLES` or something similar. Then we can sync the database by comparing the actual database schema with available Go models.

bevzzz:
@vmihailenco I've been reading through the issues and noticed this one had `help wanted` on it. I only started to work with `bun` recently, but this seems like something I could try to do. Would you like some help with it?

vmihailenco:
@bevzzz The help is welcome here, but you should be warned that this is not a small issue and it will probably require several weeks of work.

If you decide to work on this, my advice would be to learn how GORM AutoMigrate works and what we can learn from it. Writing your thoughts here so others can help would be the next step.

I did not have a chance to learn from GORM AutoMigrate yet, but I guess the plan should look like this:

1. Improve Bun dialects with ability to inspect database schema, for example, pgdialect should be capable to return a list of tables with all columns. I would start with a single PostgreSQL dialect for now.

2. Decide which operations we want to support, for example, we can start with 2 operations:

```go
package pgdialect

func (Dialect) RenameColumn(...)
func (Dialect) ChangeColumnType(...)
```

3. Use the database schema (step 1) to detect changes and apply commands from step 2 to auto-migrate the schema. Don't get too ambitious and only start with a single operation, for example, renaming a column.

Each step can be done separately so we have a chance to send feedback at the earlier stage.

Also pinging @dizzyfool and @LdDl who done some excellent work on https://github.com/dizzyfool/genna and https://github.com/LdDl/bungen and might be interested to help designing the API for step 1.

bevzzz:
@vmihailenco Thanks for getting back in touch so quickly!
Yes, you're right, bringing this feature will take some time and research, so I really appreciate the steps that you've suggested.   
I will familiarize myself with the way it's done in GORM and get back with a proposed solution for the 2 operations you've mentioned.

P.S. I plan to start working on it on the next weekend, I guess we are not in a rush here, right?

vmihailenco:
>P.S. I plan to start working on it on the next weekend, I guess we are not in a rush here, right?

No rush for sure. Even making some progress on step 1) will make a change in the long run.

LdDl:
**TLDR: just bunch of SQL queries to extract reference information about Postgres tables and my first-look thoughts**


My POV:

* Reasons to start with PostgreSQL

[...]

PR #726, opened 2022-12-06
Introduce ALTER TABLE query

bevzzz:
This PR introduces `ALTER TABLE` query as part of the effort to bring auto-migration to `bun`, see #456.
Therefore, only a [limited scope of the functionality](https://github.com/uptrace/bun/issues/456#issuecomment-1336528915) related to modifying a database schema is supported in this step.  

Changes also include:
- minor refactor of some unit tests
- some typos fixed
- tiny formatting changes (my IDE has added those, let me know if they do not fit the style and I should remove them)


bevzzz:
Something I can't make my up about is this:
When building `ALTER COLUMN SET DATA TYPE new_type`, which option should we provide?

```go
// Similarly to Column in SelectQuery, but we pass `typ` as `schema.Safe(typ)` rather than `schema.UnsafeIdent(typ)`
func (q *AlterColumnSubquery) Type(typ string) *AlterColumnSubquery {
    q.modification = schema.QueryWithArgs{
	Query: "SET DATA TYPE ?",
        Args:  []interface{}{schema.Safe(typ)},
    }
}  

// Similarly to ColumnExpr, but we append our SET DATA TYPE in front of the user's `query`
func (q *AlterColumnSubquery) TypeExpr(query string, args ...interface{}) *AlterColumnSubquery {
    	query = fmt.Sprintf("SET DATA TYPE %s", query)
	q.modification = schema.SafeQuery(query, args)
} 
```  

It is currently done in the 1st way, but I dislike it, because it is not being explicit (in the method's name) about the fact that the input is treated as a "safe" expression.  
The second way is somewhat similar in the way that `Order()` is handled in `SelectQuery`  in that it manipulates user's input before wrapping it in `schema.QueryWithArgs`, so I would be somehow leaning its way.

What do you think?

vmihailenco:
I am not sure, but we can start with collecting examples for different RDBMS.

PostgreSQL:

```sql
ALTER TABLE table_name
ALTER COLUMN column_name1 [SET DATA] TYPE new_data_type,
ALTER COLUMN column_name2 [SET DATA] TYPE new_data_type,
```

MySQL:

```sql
ALTER TABLE table_name    
MODIFY COLUMN column_name1 data_type,  
MODIFY COLUMN column_name2 data_type ... ;   
```

MSSQL:

```sql
ALTER TABLE table_name
ALTER COLUMN column_name datatype;
```

SQLite does not support it and requires creating a new table and copying existing data into it.

So to work on all supported RDBMS it should be something like:

```sql
db.NewAlterTable().AlterColumnType(columnName, columnType)
```

At least that is my line of thinking without looking at the PR.

internal/dbtest/db_test.go
```diff
@@ -224,6 +224,13 @@ func funcName(x interface{}) string {
 	return s
 }
 
+func skipIfNotHasFeature(tb testing.TB, db *bun.DB, feat feature.Feature, featName string) {
```
vmihailenco, line 227: nitpick: skipIfNoFeature

internal/dbtest/query_test.go
```diff
@@ -943,3 +945,128 @@ func TestQuery(t *testing.T) {
...
+		Old    string
+		Active bool
+	}
+
+	cases := []TestCase{
+		{
+			"rename column", needsAlterTable,
+			func(db *bun.DB) schema.QueryAppender {
+				return db.NewAlterTable().Model((*Model)(nil)).
+					RenameColumn().Column("old").To("new")
```
vmihailenco, line 962: Any reasons why not simply `RenameColumn("old_name", "new_name")`?

internal/dbtest/query_test.go
```diff
@@ -943,3 +945,128 @@ func TestQuery(t *testing.T) {
...
+			"invalid model", needsAlterTable,
+			func(db *bun.DB) schema.QueryAppender {
+				return db.NewAlterTable().Model(1)
+			},
+		},
+		{
+			"rename table", needsAlterTable,
+			func(db *bun.DB) schema.QueryAppender {
+				return db.NewAlterTable().Model((*Model)(nil)).
+					Rename().To("new_models")
```
vmihailenco, line 975: Why not `RenameTable("new_name")`?

internal/dbtest/query_test.go
```diff
@@ -943,3 +945,128 @@ func TestQuery(t *testing.T) {
...
+				return db.NewAlterTable().Model((*Model)(nil)).IfExists().
+					Rename().To("new_models")
+			},
+		},
+		{
+			"change column type", needsAlterTable,
+			func(db *bun.DB) schema.QueryAppender {
+				// Change column type with common SQL data type
+				return db.NewAlterTable().Model((*Model)(nil)).
+					AlterColumn().Column("old").Type(sqltype.Blob)
```
vmihailenco, line 990: Same suggestion `AlterColumnType(colName, colType)`

query_alter_table.go
```diff
@@ -0,0 +1,409 @@
...
+
+type SubqueryAppender interface {
+	schema.QueryAppender
+	AppendSubquery(schema.Formatter, []byte) ([]byte, error)
+}
+
+type ChainableSubquery interface {
+	SubqueryAppender
+	chain()
+}
```
vmihailenco, line 33: Looks like you've come with some new design/idea for Bun query builder. It *could be* a nice idea, but let's describe it first and come to understanding how we will use it with other queries like SELECT/UPDATE/INSERT.

query_alter_table.go
```diff
@@ -0,0 +1,409 @@
...
+
+type ifNotExists bool
+
+// AppendQuery adds IF NOT EXISTS clause to the query, if its value is true. It always returns a nil error.
+func (ifnotexists ifNotExists) AppendQuery(_ schema.Formatter, b []byte) ([]byte, error) {
+	if !ifnotexists {
+		return b, nil
+	}
+	return append(b, "IF NOT EXISTS "...), nil
+}
```
vmihailenco, line 23: Looks like over-engineering to me, but if you insist... :)

query_alter_table.go
vmihailenco: ...but I'd rather not design something new just for `ALTER TABLE` and instead try to solve the task with existing APIs.

vmihailenco:
After looking at the PR it looks like it makes sense to split `AlterTableQuery` into separate queries, e.g.:

```go
db.NewRenameTable().Table("old_table_name").To("new_table_name")
db.NewRenameTable().Model(&model).To("new_table_name")
db.NewAlterColumnType().Model(&model).Column("name1", "type1").Column("name2", "type2")
db.NewDropColumn().Model(&model).Column("col1").Column("col2")
```

WDYT?

internal/dbtest/query_test.go
bevzzz: You're right, that's actually [what I started with](https://github.com/uptrace/bun/pull/726/commits/f8ff602c226a80b3eb4d846bd6aeb062a22ae3e7), but I wasn't sure if it wasn't more consistent with the rest of the API to mirror actual the SQL keywords.  
I agree that this is much more readable and explicit though.

internal/dbtest/query_test.go
bevzzz: Ditto: https://github.com/uptrace/bun/pull/726#discussion_r1044640112

internal/dbtest/query_test.go
bevzzz: Ditto: https://github.com/uptrace/bun/pull/726#discussion_r1044640112

bevzzz:
> That would make API cleaner and we won't need chainable/parent queries any more. WDYT?

**tldr;**
Yes, I definitely agree that it would make sense to handle each modification separately, primarily because of the stark differences in syntax and degree of support for `ALTER TABLE` in the 4 DBMS.

**Syntax and feature support**
> I am not sure, but we can start with collecting examples for different RDBMS.

Following your suggestion, I've put up the examples for most of the queries from [this list](https://github.com/uptrace/bun/issues/456#issuecomment-1336528915) in the dialects Bun supports.
I'm posting them here as a convenient summary and also as a reference for future development:

<details>  
<summary>RENAME TABLE</summary>  

```sql
-- PostgresSQL
ALTER TABLE t RENAME TO new;

-- MySQL
ALTER TABLE t RENAME TO new;
-- or, for multiple renamed tables:
RENAME TABLE
	t TO new,
	d TO dew;

-- MSSQL
EXEC sp_rename 't', 'new', 'TABLE';

-- SQLite
ALTER TABLE t RENAME TO new;
```
</details>


<details>  
<summary>RENAME COLUMN</summary>  

```sql
-- PostgresSQL
ALTER TABLE t RENAME COLUMN a TO new_a;

-- MySQL
ALTER TABLE t RENAME COLUMN a TO new_a;

-- MSSQL
EXEC sp_rename 't.a', 'col_a', 'COLUMN'

-- SQLite
ALTER TABLE t RENAME COLUMN a TO new_a;
```
</details>


<details>  
<summary>CHANGE COLUMN TYPE</summary>  

```sql
-- PostgresSQL
ALTER TABLE t
	ALTER COLUMN a SET DATA TYPE new_data_type,
	ALTER COLUMN b SET DATA TYPE new_data_type;

-- MySQL
ALTER TABLE t
	MODIFY a new_data_type,
	MODIFY b new_data_type;

-- MSSQL
ALTER TABLE table_name ALTER COLUMN a new_data_type;
ALTER TABLE table_name ALTER COLUMN b new_data_type;


-- SQLite
Not supported in the standard API, probably something like this:
BEGIN TRANSACTION
ALTER TABLE t ADD COLUMN __temp_a new_data_type_a;
UPDATE t SET __temp_a = a;
ALTER TABLE t DROP COLUMN a;
ALTER TABLE t RENAME COLUMN __temp_a TO a;
COMMIT;
```
</details>  

<details>  
<summary>ADD COLUMN</summary>  

```sql
-- PostgresSQL
ALTER TABLE t
	ADD COLUMN a data_type,
	ADD COLUMN b data_type;

-- MySQL
ALTER TABLE t
	ADD COLUMN a data_type,
	ADD COLUMN b data_type;

-- MSSQL
ALTER TABLE t
ADD 
	a data_type,
	b data_type;

-- SQLite
ALTER TABLE t ADD COLUMN a data_type;
ALTER TABLE t ADD COLUMN b data_type;
```
</details>  


<details>  
<summary>DROP COLUMN</summary>  

```sql
-- PostgresSQL
ALTER TABLE t 
	ALTER COLUMN a SET DEFAULT 0,
	ALTER COLUMN b SET DEFAULT 0;

-- MySQL
ALTER TABLE t
	ALTER COLUMN a SET DEFAULT 0,
	ALTER COLUMN b SET DEFAULT 0;

-- MSSQL
ALTER TABLE t ADD
	CONSTRAINT my_default_value_a DEFAULT 0 FOR a,
	CONSTRAINT my_default_value_a DEFAULT 0 FOR b; 

-- SQLite
Not supported in the standard API:

BEGIN TRANSACTION
ALTER TABLE t ADD COLUMN __temp_a data_type_a DEFAULT 0;
UPDATE t SET __temp_a = a;
ALTER TABLE t DROP COLUMN a;
ALTER TABLE t RENAME COLUMN __temp_a TO a;
COMMIT;
```
</details>    


There are 2 things to notice here:
- _SQLite has limited support for ALTER TABLE_. Rename table, rename/add/drop column are the only 4 queries supported, the rest must be done by moving data to temp columns/tables and back.
- _Some DMBS allow executing multiple statements within one query_. This was the reason I've introduced the `ChainableSubquery` in the first place -- that should've been a mechanism for e.g., changing data types in 5 columns, adding a new column, and dropping a default constraint from another one within a single SQL statement. However, with that not being the case for all 4 DBMS (PostgresSQL and MySQL the most consistent support for "chainable subqueries"), I think it's not worth breaking our backs trying to accommodate that in the Bun's API in this iteration. 
- _Different "angles" _. For example, while setting a default is "altering the column definition" for PostgreSQL and MySQL, it is "adding a table constraint" in MSSQL. 

In conclusion, I think these APIs for altering DB schema have less commonalities than I have initially hoped for, and in this case it really makes more sense to design Bun's API around _specific operations_ rather than _SQL syntactic constructs_.

query_alter_table.go
bevzzz: Agree.  
I've added some more thoughts about that [below](https://github.com/uptrace/bun/pull/726#issuecomment-1344649186).  

Going to refactor `ChainableSubquery` away in favor of the existing API :ok_hand:.  
It can stay in Git history, or I could describe it in a proposal, if you think it could be used in the rest of the lib.

vmihailenco:
>really makes more sense to design Bun's API around specific operations rather than SQL syntactic constructs.

Fully agree. Perhaps even ignore `AlterTableQuery` altogether and just provide minimal API like:

```go
RenameTable(context.Context, oldName, newName string) error
AddColumn(context.Context, *sqlschema.Column) error
DropColumn(context.Context, colName string) error
ChangeColumnType(context.Context, colName, colType string) error
```

This way you can do whatever you want in dialects, .e.g. start transactions, create temp tables etc.

bevzzz:
After giving it some thought here's roughly the solution I came up with:
1. Define `AlterTableQuery`, which has methods for individual table-modifications
```go
db.NewAlterTable().Model(...).RenameTable("new_name")
db.NewAlterTable().Model(...).AddColumn("col_2", sqltype.BigInt)
```  

The reason I want to continue with the standard `bun`-query approach is:
a) familiar syntax
b) the way models are parsed in `baseQuery` is very helpful and easily reusable

2. Delegate building the query to individual dialects. Because of the differences in the syntax, they seem like the best place to handle that
```go
type Dialect interface {
    RenameTable(...) []QueryWithArgs
    AddColumn(...)   []QueryWithArgs
}
```
These methods return a slice of queries, so that, for instance, SQLite could build _several queries_ for something like `SET DEFAULT` (which is not natively supported). 

3. Execute multiple raw queries in a transaction. The "exec" step belongs to `bun`, because that's the part of the program that holds the db connection.
```go
func (q *AlterTableQuery) Exec(ctx) (sql.Result, error) {
    if len(q.mod) > 1 {
        // execute raw query at q.mod[0]
    }
    // start a transaction and execute each one of the queries in q.mod
}
```

Going to add implementation for that tomorrow, let's see how that works out.

bevzzz:
You're right, I too thought that returning `[]QueryWithArgs`  felt awkward. The reason I would like the dialects to only build the query (doesn't have to be a "bun query") itself and to handle its execution in a separate package is something I've mentioned in one of my [earlier comments](https://github.com/uptrace/bun/issues/456#issuecomment-1335177025):

> Once we've got auto-migrations figured out, we could leverage the fact that all bun's queries implement Stringer and create "up" and "down" SQL-migration files for the user.

The only alternative that comes to my mind atm is to have the dialects return `[]byte` like other `QueryAppender`s do, but it's only a slight improvement over `[]QueryWithArgs`. WDYT? Is there a more elegant solution?


> I guess you are worried about not having access to bun.DB in dialects, but we can introduce a separate sub-package that works with bun.DB and does some dispatching for different dialects internally...

Yes, if we put `ALTER TABLE` functionality in a sub-separate package, we could just pass `bun.DB` as an argument.

vmihailenco:
>build the query (doesn't have to be a "bun query") itself and to handle its execution in a separate package

I've missed that, but it would be nice to have.

Perhaps we could achieve that by having some special `*sql.DB` abstracted under an interface that would record queries instead of executing them. For example, https://github.com/DATA-DOG/go-sqlmock records queries for testing, but we could write them into a file instead.

PR #729, opened 2022-12-07
feat: Support for plugins based on *sql.DB

goferHiro:
I realized that sql Compatible plugins like https://github.com/goferHiro/dbresolver can be supported by reducing the dependency of bun into database/sql.

Note that *sql.DB is just a struct with a couple of private fields and exported methods. Having an interface in its place has no additional costs in terms of readability or gc

Signed-off-by: Hiro <goferHiro@gmail.com>


db.go
```diff
@@ -32,8 +34,31 @@ func WithDiscardUnknownColumns() DBOption {
 	}
 }
 
+type SQLRepo interface {
```
vmihailenco, line 37: What is SQLRepo? It sounds strange... Perhaps `SQLDB`?

vmihailenco:
Please fix the build.

db.go
goferHiro: https://threedots.tech/post/repository-pattern-in-go/

Ideally this is to show that it is a  Repo pattern followed.

Initially I had it as SQLDB but found harder to read since all letters are caps

goferHiro: Plus the whole point was to give an independency for developers  from database/sql, so that dbresolvers, dbloggers based on database/sql can be integrated to bun

goferHiro:
> Please fix the build.

Can u help me realize how to build? I know how to build the hard way i.e by refactoring all the uptrace pkgs to my repo. 

Is there another way?  Can u give me contributor access?

goferHiro:
@vmihailenco  Can we connect via slack?

https://join.slack.com/t/bxcodecoss/shared_invite/zt-1l8qt2cie-zlTlYSw3oSINoOY3fekYIw

goferHiro:
failed.

Issue #611, opened 2022-07-10
Bug: pgdriver.WithConnParams search_path does not allow multiple parameters/schemas


tomalaci:
The issue #214 and its PR #220 added ability to change search_path, however, it seems to only allow single schema/parameter. If you specify comma-separated string of multiple schemas then it constructs wrong kind of SET statement.

For example, suppose you have a schema "myschema" where you plan to have some tables and then you want to have a common schema "extensions" to store extensions in. A scenario described in this StackOverflow answer: https://stackoverflow.com/a/43937189

The above scenario should have following search_path set:
```sql
SET search_path = "myschema", "extensions";
```

Unfortunately bun's pgdriver.WithConnParams formats parameters as a single string. So this...
```go
pgdriver.WithConnParams(map[string]interface{}{
	"search_path": "myschema, extensions",
})
```
... becomes this ...
```sql
SET search_path = "myschema, extensions";
```
... and that doesn't seem to work for correctly finding common extensions.

I also tried specifying search_path parameters with `[]string` type but it failed with `pgdriver: unexpected arg: []string`.

**Suggestion for solution**

You should allow `[]string` values in `pgdriver.WithConnParams`. So specifying this:
```go
pgdriver.WithConnParams(map[string]interface{}{
	"search_path": []string{"myschema", "extensions"}
})
```
... would properly separate each string element into its own quoted element in SET statement like this ...
```sql
SET search_path = "myschema", "extensions";
```

Let me know if something is unclear! Thanks!

PR #732, opened 2022-12-26
Allow string array in WithConnParams

mik-volkov:
#611 
This PR allows multiple parameters to WithConnParams in the form of string slice
```
		pgdriver.WithConnParams(map[string]interface{}{
			"search_path": []string{"foo", "bar"},
		}),
```


vmihailenco:
This PR assumes that `[]string{"foo", "bar"}` can be appended as is and does not require any quotation which might not be the case.

mik-volkov:
Added quotation to parameters processing

qindj:
could you please merge this? I met the same issue, thanks

PR #733, opened 2022-12-26
chore: introducing go workspaces

goferHiro:
go run github.com/uptrace/bun/example/basic@latest fails due to replace directives in go.mod. A clean solution to this is go workspaces.


vmihailenco:
@goferHiro do we need go 1.19 to use this? Looks like go workspaces are available since Go 1.18...

goferHiro:
true. go 1.19 is faster in terms of compilation speeds. I saw this

image


Hence presumed, you are looking forwards towards upgrading go. 

If its going to be a problem, i ill change the go version to 1.18 its just one regex away!

vmihailenco:
>If its going to be a problem, i ill change the go version to 1.18 its just one regex away!

Yes, please change to Go 1.18. We aim to support last 2 Go version and Go 1.20 is not released yet.

goferHiro:
Done

goferHiro:
@vmihailenco I am open to working in any issues. Just let me know

goferHiro:
@vmihailenco  i would like to know your opinion on this https://github.com/uptrace/bun/issues/735

vmihailenco:
Thanks :+1:

PR #737, opened 2022-12-28
feat: add support for time.Time array in Postgres

jhhom:
Using `[]time.Time` with `bun:",array"` causes SQL syntax error when inserting into PostgreSQL.

The syntax error is caused by the single quotes surrounding the array, e.g `'{'2022-10-31 23:21:49.673242+00:00'}'`.

This PR fixes that by providing a custom formatter for PostgreSQL dialect that uses double quotes to surround the time instead, e.g `'{"2022-10-31 23:21:49.673242+00:00"}'`


vmihailenco:
Thanks :+1:

Issue #736, opened 2022-12-28
Bug: CreateTableQuery fails to set VARCHAR length (MSSQL and MySQL)


bevzzz:
I've encountered this odd behavior with `CreateTableQuery.Varchar(n)`:

```go
// Regular query
db.NewCreateTable().Model(new(Model))
// Output:
// pg:			CREATE TABLE "models" ("id" VARCHAR)
// pgx:			CREATE TABLE "models" ("id" VARCHAR)
// sqlite:		CREATE TABLE "models" ("id" VARCHAR)
// mysql5:		CREATE TABLE `models` (`id` VARCHAR(255))
// mysql8:		CREATE TABLE `models` (`id` VARCHAR(255))
// mariadb:		CREATE TABLE `models` (`id` VARCHAR(255))
// mssql2019: 	        CREATE TABLE "models" ("id" VARCHAR(255))

// Specify VARCHAR length
db.NewCreateTable().Model(new(Model)).Varchar(10)
// Output:
// pg:        CREATE TABLE "models" ("id" varchar(10))
// pgx:       CREATE TABLE "models" ("id" varchar(10))
// sqlite:    CREATE TABLE "models" ("id" varchar(10))
// mysql5:    CREATE TABLE `models` (`id` VARCHAR(255))
// mysql8:    CREATE TABLE `models` (`id` VARCHAR(255))
// mariadb:   CREATE TABLE `models` (`id` VARCHAR(255))
// mssql2019: CREATE TABLE "models" ("id" VARCHAR(255))
```
I've also noticed there is no unit tests that check `.Varchar()` method, so I guess it could be a bug.

**Problem**
Specifying `VARCHAR` length is mandatory in MySQL and MSSQL, so the default 255 is hardcoded in their `DiscoveredSQLType`:

https://github.com/uptrace/bun/blob/733affdc3f5f30ba3ad93a5b2808012a4c8497c1/dialect/mssqldialect/dialect.go#L133

https://github.com/uptrace/bun/blob/733affdc3f5f30ba3ad93a5b2808012a4c8497c1/dialect/mysqldialect/dialect.go#L178

As a result, [this function](https://github.com/uptrace/bun/blob/a95dfffa1f03c2e176f80aae872cbe421db0f6a1/query_table_create.go#L223-L237) in `query_table_create.go` fails to recognize `VARCHAR` column in those dialects :
<details>    
<summary>Click for a commented version</summary>  

```go
[...]

PR #738, opened 2022-12-28
CreateTableQuery: use correct VARCHAR length

bevzzz:
This PR fixes the handling of VARCHAR length in `CREATE TABLE`, see #736. 

Changes:
- `Dialect` implementations now need to have `DefaultVarcharLen() uint`, so that we wouldn't need to hardcode the default lengths in the dialect-specific types
- Minor refactor of `CreateAlterTable.appendSQLType`
- ~~Since `VARCHAR` can only have positive length, the signature of `CreateAlterTable.Varchar` is changed to `Varchar(n uint)`.
[This commit](df74d485f335f8a38b29b6d5d9b172513872db0f) , therefore, has a BREAKING CHANGE ❗, as this would no longer compile:~~

```golang
var n int = 4
db.NewCreateTable().Model(...).Varchar(n)
```  

Edit: breaking changes were reverted, see [this commit](https://github.com/uptrace/bun/pull/738/commits/3335e0b9d6d3f424145e1f715223a0fffe773d9a) instead.

This change is not necessary for the fix, I only propose this for the sake of additional "type safety" in the queries. It feels more transparent than something like `n = max(0, n)`. However, if such API change is not acceptable, we could also return an error for `q.varchar <= 0` or just revert this commit entirely.

### DIFF
--- a/dialect/mssqldialect/dialect.go
+++ b/dialect/mssqldialect/dialect.go
@@ -127,7 +127,7 @@ func (*Dialect) AppendBool(b []byte, v bool) []byte {
 	return strconv.AppendUint(b, uint64(num), 10)
 }
 
-func (d *Dialect) DefaultVarcharLen() int {
+func (d *Dialect) DefaultVarcharLen() uint {
 	return 255
 }
 
--- a/dialect/mssqldialect/dialect.go
+++ b/dialect/mssqldialect/dialect.go
@@ -127,10 +127,12 @@ func (*Dialect) AppendBool(b []byte, v bool) []byte {
 	return strconv.AppendUint(b, uint64(num), 10)
 }
 
+func (d *Dialect) DefaultVarcharLen() int {
+	return 255
+}
+
 func sqlType(field *schema.Field) string {
 	switch field.DiscoveredSQLType {
-	case sqltype.VarChar:
-		return field.DiscoveredSQLType + "(255)"
 	case sqltype.Timestamp:
 		return datetimeType
 	case sqltype.Boolean:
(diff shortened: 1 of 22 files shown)


vmihailenco:
@bevzzz looks good, thanks.

Could you clarify what happens when a dialect returns `varchar=0` and `CreateTableQuery` does not specify `Varchar` length either? Do we have a test in query_test for such a case?

>However, if such API change is not acceptable, we could also return an error for q.varchar <= 0 or just revert this commit entirely.

Let's not break the existing API. Instead, let's return an error like you suggested or just `panic` if returning an error is complicated.

Also, `len(slice)` can return only positive values too, but the return type is `int`, not `uint`. Just saying :)

bevzzz:
> what happens when a dialect returns varchar=0 and CreateTableQuery does not specify Varchar length either?

We initialize `q.varchar` with the dialect's default length (see [here](https://github.com/uptrace/bun/pull/738/files#diff-7fc334b1ed7219952ece271041041ad0ad5128ffeaf3bf6902655bd85288a923R41)), so if no other `Varchar` is specified (`q.varchar == 0`), we do not append anything in the parenthesis and the type stays `VARCHAR` . 

Yes, there's a test case for that:
```go
func(db *bun.DB) schema.QueryAppender {
    return db.NewCreateTable().Model(new(Model))
}
```  
It's one of the older ones (no regressions ✌️ ), and for PG, which has default 0, it generates [this query](https://github.com/uptrace/bun/blob/master/internal/dbtest/testdata/snapshots/TestQuery-pg-32):
```
CREATE TABLE "models" ("id" BIGSERIAL NOT NULL, "str" VARCHAR, PRIMARY KEY ("id"))
```  

> Let's not break the existing API  

Agreed. Good point about `len(slice)` 👍

vmihailenco:
Thanks :+1:

PR #774, opened 2023-02-02
feat(feat-when): bun add when method like laravel query builder

xiusin:
**Make the code more elegant** 

```go
if form.MediaId > 0 {
	query.Where("media_id = ?", form.MediaId)
}
```
to 

```go
query.When(form.MediaId > 0, func(query *bun.SelectQuery) {
   query.Where("media_id = ?", form.MediaId);
})
```

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -139,6 +139,13 @@ func (q *SelectQuery) WherePK(cols ...string) *SelectQuery {
 	return q
 }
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
The code becomes more elegant I guess, but it also requires studying what method `When` does which is something you don't need to do with the good old `if`.

Thanks for the effort anyway.

xiusin:
No, I just think that's better. tks

Issue #776, opened 2023-02-06
panic: reflect: call of reflect.Value.Field on ptr Value on v1.1.10


Link512:
The following code snippet worked on `v1.1.9` and started panicking on `v1.1.10`. 

```go
package main

import (
	"context"
	"database/sql"

	"github.com/uptrace/bun"
	"github.com/uptrace/bun/dialect/pgdialect"
	"github.com/uptrace/bun/driver/pgdriver"
)

type Model struct {
	bun.BaseModel `bun:"users"`

	Name     string  `bun:"name"`
	Nickname *string `bun:"nickname"`
}

func main() {
	ctx := context.Background()
	driver := pgdriver.NewConnector(pgdriver.WithDSN("postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable"))
	db := bun.NewDB(sql.OpenDB(driver), pgdialect.New())

	db.NewCreateTable().Model((*Model)(nil)).IfNotExists().Exec(context.Background())

	m := &Model{Name: "bob"}
	models := []*Model{m}

	db.NewInsert().Model(&models).Exec(ctx) // <- panics because only `nickname` is being returned and is NULL

	res := []*Model{}
	db.NewSelect().Model(&res).Scan(ctx)                    // <- doesn't panic
	db.NewSelect().Model(&res).Column("nickname").Scan(ctx) // <- panics becase we only want `nickname`, which is NULL

	db.NewDropTable().Model((*Model)(nil)).IfExists().Exec(context.Background())
}
```

This issue can be reproduced in scenarios where:
- you use slices of model pointers and
- you either `SELECT` or read `RETURN`ed columns that are *all* `NULL`.

After some debugging I belive I have tracked down the issue to [this commit](https://github.com/uptrace/bun/commit/67e4412a972a9ed5f3a1d07c66957beedbc8a8a3) :
 1. `.Elem()` was removed from the return value
 2. `sliceTableModel.ScanRows()` calls [this function](https://github.com/uptrace/bun/blob/master/model_table_slice.go#L70) and `m.strct` is a `reflect.Pointer` instead of a `reflect.Struct`
 3. `structTableModel.scanColumn` calls [this function](https://github.com/uptrace/bun/blob/master/model_table_struct.go#L331), which will subsequently call `strct.Field(index)`, causing the `reflect` package to panic, since `strct` is not a Struct type
[...]

PR #777, opened 2023-02-06
add test for insert panics

Link512:
Unit test for easy reproducing of panic described in the issue https://github.com/uptrace/bun/issues/776

### DIFF
--- a/model_table_struct.go
+++ b/model_table_struct.go
@@ -74,15 +74,11 @@ func (m *structTableModel) initStruct() error {
 		return m.structInitErr
 	case reflect.Interface:
 		m.strct = m.strct.Elem()
-	}
-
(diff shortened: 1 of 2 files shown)


model_table_struct.go
```diff
@@ -74,15 +74,11 @@ func (m *structTableModel) initStruct() error {
 		return m.structInitErr
 	case reflect.Interface:
 		m.strct = m.strct.Elem()
-	}
-
-	if m.strct.Kind() == reflect.Ptr {
+	case reflect.Ptr:
```
vmihailenco, line 77: This is not safe, because on line 76 we do `m.strct = m.strct.Elem()` which might return a nil ptr.

PR #788, opened 2023-02-28
Expose Exec(…) method for RawQuery

l0nax:
The Exec(…) method allows to execute raw queries which do not return any data.

This is required in order to run queries like the following one:
```go
	q2 := db.NewRaw(
		`SELECT
			id,
			title
		INTO TEMPORARY
			tmp_stories_title
		FROM
			stories
		`)

	_, err = q2.Exec(ctx)
	if err != nil {
		panic(err)
	}
```

Or to `REFRESH` a materialized view.


vmihailenco:
Thanks :+1:

PR #796, opened 2023-03-05
feat: add option to customize Go migration template

greeflas:
Discussion #795

### DIFF
--- a/migrate/migrator.go
+++ b/migrate/migrator.go
@@ -217,6 +217,7 @@ func (m *Migrator) Rollback(ctx context.Context, opts ...MigrationOption) (*Migr
 
 type goMigrationConfig struct {
 	packageName string
+	content     string
 }
 
 type GoMigrationOption func(cfg *goMigrationConfig)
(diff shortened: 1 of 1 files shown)


migrate/migrator.go
```diff
@@ -227,12 +228,19 @@ func WithPackageName(name string) GoMigrationOption {
 	}
 }
 
+func WithContent(content string) GoMigrationOption {
```
vmihailenco, line 231: Please rename to `WithGoTemplate` so we can also have `WithSQLTemplate` later.

migrate/migrator.go
```diff
@@ -217,6 +217,7 @@ func (m *Migrator) Rollback(ctx context.Context, opts ...MigrationOption) (*Migr
 
 type goMigrationConfig struct {
 	packageName string
+	content     string
```
vmihailenco, line 220: content -> goTemplate

migrate/migrator.go
greeflas: Done.

migrate/migrator.go
greeflas: Done.

vmihailenco:
Thanks :+1: Ideally, we should also start using text/template here to make code generation more flexible. Guess this can done in a separate PR later - https://github.com/uptrace/bun/issues/797

Issue #822, opened 2023-05-03
MySQL: DELETE table aliases don't work in version older than 8.0.16


jsclayton:
We ran across an error today when deploying Bun against an Azure MySQL database and found that table aliases in `DELETE` statements are only supported in v8.0.16 and higher. It appears that Bun's MySQL dialect considers that feature available in any version v8.0 or higher.

From the bottom of the [MySQL 8.0 `DELETE` documentation](https://dev.mysql.com/doc/refman/8.0/en/delete.html):

```
Table aliases are also supported for single-table DELETE statements beginning with MySQL 8.0.16. (Bug #89410,Bug #27455809)
```

PR #823, opened 2023-05-03
fix: Allow MySQL table aliases for DELETE in >= v8.0.16

jsclayton:
Fixes https://github.com/uptrace/bun/issues/822


vmihailenco:
Thanks

PR #831, opened 2023-05-18
fix: enable CompositeIn for MySQL

calorie:
close: https://github.com/uptrace/bun/pull/632

MySQL support `IN` conditions with multiple fields.

See https://dev.mysql.com/doc/refman/5.7/en/comparison-operators.html#operator_in

So, I added `feature.CompositeIn` to the features.


yu-shigetani-cover:
@vmihailenco 
Could you review it ?

vmihailenco (approved):

PR #836, opened 2023-05-26
feat(bunotel): add options for set otel providers of `bunotel.QueryHook`

Tiscs:
Allow using a specified TracerProvider and MeterProvider instead of the global ones.

- usage:
  ``` go
  db := bun.NewDB(sqldb, dialect, opts...)
  var tp trace.TracerProvider
  var mp metric.MeterProvider
  // initialize tracer and meter provider
  db.AddQueryHook(bunotel.NewQueryHook(bunotel.WithTracerProvider(tp), bunotel.WithMeterProvider(mp)))
  ```

### DIFF
--- a/extra/bunotel/option.go
+++ b/extra/bunotel/option.go
@@ -2,7 +2,9 @@ package bunotel
 
 import (
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
 	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
+	"go.opentelemetry.io/otel/trace"
 )
 
 type Option func(h *QueryHook)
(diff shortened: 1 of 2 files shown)


extra/bunotel/option.go
```diff
@@ -30,3 +32,28 @@ func WithFormattedQueries(format bool) Option {
...
+	}
+}
+
+// WithMeterProvider returns an Option to use the MeterProvider when
+// creating a Meter.
+func WithMeterProvider(mp metric.MeterProvider) Option {
+	return func(h *QueryHook) {
+		if mp != nil {
+			h.meter = mp.Meter("github.com/uptrace/bun")
+			h.queryHistogram, _ = h.meter.Int64Histogram(
```
vmihailenco, line 52: Looks like this will be overridden anyway later when the hooks is created.

vmihailenco: So, let's just remove this?

extra/bunotel/option.go
```diff
@@ -30,3 +32,28 @@ func WithFormattedQueries(format bool) Option {
 		h.formatQueries = format
 	}
 }
+
+// WithTracerProvider returns an Option to use the TracerProvider when
+// creating a Tracer.
+func WithTracerProvider(tp trace.TracerProvider) Option {
+	return func(h *QueryHook) {
+		if tp != nil {
+			h.tracer = tp.Tracer("github.com/uptrace/bun")
```
vmihailenco, line 41: It is a bit strange that you pass `trace.TracerProvider` just to create a tracer. Also, what if you want to change tracer name?

Can't we instead accept a `tracer`?

The same applies to `WithMeterProvider`. Should not we accept a `meter`?

extra/bunotel/option.go
Tiscs: You are right, I forgot to remove this.

extra/bunotel/option.go
Tiscs: I considered directly configuring the tracer and meter through options instead of TracerProvider and MeterProvider. However, I decided to do it this way because in the future, there might be a need to create different tracers and meters within the QueryHook. The current approach ensures compatibility in the future without breaking any existing functionality.
Thanks.

Tiscs: Moreover, I think it is advisable to avoid allowing users to customize the name of tracer and meter to prevent ambiguity in the otel information.

Tiscs:
Hi, any responses here?

vmihailenco:
Looks good, thanks!

PR #848, opened 2023-06-05
fix: :bug: apply relation functions on count queries

### LINKED ISSUE #597: Bug : functions in Relation statements are not applied on select queries terminated with Count
Hello,

I found a weird bug with the `apply` function that can be passed to the `Relation` method of a `SelectQuery` when it is coupled with the `Count` query termination.

With the following models.
```go
type A struct {
   id   int64 `bun:",pk,autoincrement"`
   name string 
}

type B struct {
   id  int64 `bun:",pk,autoincrement"`
   aId int64
   a   *A    `bun:"rel:belongs-to,join:aId=id"`
}
```

The `WHERE` clause in the query bellow will be ignored.
```go
count, err := db.NewSelect().
   Model(&[]*B{}).
   Relation("A", func(q *bun.SelectQuery) *bun.SelectQuery {
      return q.Where("name = ?", "someName")
   }).
   Count(ctx)
```

However this query works as expected.
```go
err := db.NewSelect().
   Model(&[]*B{}).
   Relation("A", func(q *bun.SelectQuery) *bun.SelectQuery {
      return q.Where("name = ?", "someName")
   }).
   Scan(ctx)
```

enzalito:
fixes #597

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -923,6 +923,17 @@ func (q *SelectQuery) Count(ctx context.Context) (int, error) {
 		return 0, q.err
 	}
 
+	if err := q.forEachInlineRelJoin(func(j *relationJoin) error {
+		j.applyTo(q)
+		return nil
+	}); err != nil {
+		return 0, err
+	}
+
+	return q.count(ctx)
+}
+
+func (q *SelectQuery) count(ctx context.Context) (int, error) {
 	qq := countQuery{q}
 
 	queryBytes, err := qq.AppendQuery(q.db.fmter, nil)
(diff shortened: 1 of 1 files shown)


query_select.go
```diff
@@ -923,6 +923,17 @@ func (q *SelectQuery) Count(ctx context.Context) (int, error) {
 		return 0, q.err
 	}
 
+	if err := q.forEachInlineRelJoin(func(j *relationJoin) error {
+		j.applyTo(q)
```
vmihailenco, line 927: It looks like this function modifies the `q`. Meaning that it will be applied twice when run with `ScanAndCount`:

- first it will be applied by Scan
- second it will be applied by Count

If that is true, then the query generation becomes racy and unsafe....

enzalito: ScanAndCount calls the unexported `count` function that corresponds to `Count` on the main branch. If I remember correctly the lines you mention do in fact modify `q` but they are only run when `Count` (not `count`) is called directly.  

I made this PR a while ago so I might be missing something, you tell me

PR #849, opened 2023-06-05
fix: :bug: look into the fields of the joins during has-many models field resolution

enzalito:
fixes #847

Issue #847, opened 2023-06-05
Error `bun: TableName does not have column=ColumnName"` on relations nested inside a has-many relation


enzalito:
I found a bug that seems to happen with queries that include a has-many relation and nested relations that refer to the same table (not very clear I know, refer to the reproducer bellow). It took me some time to pinpoint the cause of the problem but I eventually found it. I will attach a PR to this issue with the fix I came up with.

```go
package main

import (
	"context"
	"database/sql"
	"encoding/json"
	"fmt"

	embeddedpostgres "github.com/fergusstrange/embedded-postgres"
	"github.com/uptrace/bun"
	"github.com/uptrace/bun/dialect/pgdialect"
	"github.com/uptrace/bun/driver/pgdriver"
)

type Foo struct {
	Id   int64 `bun:",pk,autoincrement,nullzero"`
	Name string
	Bars []*Bar `bun:"rel:has-many,join:id=foo_id"`
}

type Bar struct {
	Id    int64 `bun:",pk,nullzero"`
	Name  string
	FooId int64 `bun:",nullzero"`
	BazId int64 `bun:",nullzero"`
	Baz   *Baz  `bun:"rel:belongs-to,join:baz_id=id"`
	Quux  *Quux `bun:"rel:has-one,join:id=id"`
}

type Baz struct {
	Id    int64 `bun:",pk,nullzero"`
	Name  string
	QuxId int64 `bun:",nullzero"`
	Qux   *Qux  `bun:"rel:belongs-to,join:qux_id=id"`
	Quux  *Quux `bun:"rel:has-one,join:id=id"`
}

type Qux struct {
	Id   int64 `bun:",pk,nullzero"`
	Name string
	Quux *Quux `bun:"rel:has-one,join:id=id"`
}

type Quux struct {
	Id   int64 `bun:",pk,autoincrement,nullzero"`
	Name string
}

func main() {
	pgServer := embeddedpostgres.NewDatabase(
		embeddedpostgres.DefaultConfig().Logger(nil),
	)
	err := pgServer.Start()
	if err != nil {
		panic(err)
	}
	defer pgServer.Stop()

	ctx := context.Background()
	dsn := "postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable"
	sqlDb := sql.OpenDB(pgdriver.NewConnector(pgdriver.WithDSN(dsn)))
	db := bun.NewDB(sqlDb, pgdialect.New())

	_, err = db.NewCreateTable().
[...]

PR #850, opened 2023-06-06
fix "model does not have column" error

martoche:
The following type has two belongs-to fields with different names (BillingAddress and ShippingAddress) but with the same type (Address):

```
	type CompanyAddress struct {
		Id                int `bun:",pk"`
		CompanyId         int
		BillingAddressId  int
		BillingAddress    *Address `bun:"rel:belongs-to,join:billing_address_id=id"`
		ShippingAddressId int
		ShippingAddress   *Address `bun:"rel:belongs-to,join:shipping_address_id=id"`
	}
```

It causes this error when making a `SELECT` (see test case):

```
sql: Scan error on column index 7, name "company_address__shipping_address__id": bun: model=Company does not have column=company_address__shipping_address__id
```

I've attached a test case to demonstrate the problem.

This commit is an attempt to fix the bug. I've narrowed it down to the [Table.inlineFields](https://github.com/uptrace/bun/blob/dacffd2ac21ad0be76edb0425b806696728a29e1/schema/table.go#LL809C17-L809C29) method.

There is a `seen` map that saves which fields have already been visited to avoid an infinite loop in case of circular relations. In the case with two fields with different names but the same type (BillingAddress and ShippingAddress), the `seen` map prevents the `inlineFields` method to visit one of the two fields, which causes the error.

This is just one attempt to fix the bug, I'm happy to make changes if you think it should be fixes in another way.

(Another attempt by another contributor to fix this bug has been made in #849, original bug #847. This other PR doesn't have a test case, but fixes the problem elsewhere in the code, it might be interesting to compare the two different fixes.)

### DIFF
--- a/internal/dbtest/orm_test.go
+++ b/internal/dbtest/orm_test.go
@@ -31,7 +31,9 @@ func TestORM(t *testing.T) {
 		{testRelationExcludeAll},
 		{testM2MRelationExcludeColumn},
 		{testRelationBelongsToSelf},
+		{testRelationsCycle},
 		{testCompositeHasMany},
+		{testRelationsDifferentFieldsWithSameType},
 	}
 
 	testEachDB(t, func(t *testing.T, dbName string, db *bun.DB) {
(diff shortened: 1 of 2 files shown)


TonDar0n:
Hi, thank you for this PR.  

I wanted to inform you that there is a flaw in the way you're checking if a field has already been visited.
With nested relations, fields with the same type and the same name can be involved in a request. In that case, the bug will still happen.
You can use the reproducer I gave in https://github.com/uptrace/bun/issues/847 to see for yourself.

Storing the index of the field in `inlineFieldsSeenKey` instead of its name should fix the problem. It is a slice though so, if you choose this approach, stringifying it will be necessary.

martoche:
Thanks for your input @TonDar0n, I've integrated your test case and fixed the code accordingly.

martoche:
I've just force-pushed a new version, adding an new test case for circular relations.

The `seen` map was previously shared across all calls to `inlineFields`, stopping the recursion as soon as a field with a previously seen type was visited.

Now, the seen map is cloned so that all the paths of what has been visited is saved, and we're able to stop the recursion more precisely: only when a path starts looping.

@vmihailenco this is ready for review, thanks!

vmihailenco:
Merging since this is an improvement to the current behavior, but this does not look like a complete fix for the recursion case. We just give up at some later point then before.

https://github.com/uptrace/bun/pull/849 looks more promising since it adds dynamic resolving, but it requires more work...

vmihailenco:
This is reverted by https://github.com/uptrace/bun/commit/387228e85d22dfcf3659f4631dfa87106d7ef45f due to https://github.com/uptrace/bun/pull/897

PR #851, opened 2023-06-13
fix: Index hints are not working

calorie:
Index hints have to be specified following a table name. 

See: https://dev.mysql.com/doc/refman/8.0/en/index-hints.html


vmihailenco:
THanks

Issue #858, opened 2023-06-28
Bug Report: Update with empty map/struct model causes invalid SQL generation


jamesjulich:
If you ask bun to create an update statement using a model with all nil fields, then call OmitZero, you will end up with an invalid query as seen here.

```
db.NewUpdate().
	Where("? = ?", bun.Ident("id"), c.Params("id")).
	OmitZero().
	Model(&myEmptyStruct{}). // also confirmed to happen with empty map, ie. new(map[string]interface{})
	String();
```

will produce SQL invalid SQL similar to this: ```UPDATE "test" AS "test_struct" SET  WHERE ("id" = '1')```

I think #684 is reporting the same issue, but it's been dormant for a while and I think this warrants bumping.

Since 'SET' is a required field, I think it makes sense to raise an error if nothing is appended to the SET statement in [mustAppendSet](https://github.com/uptrace/bun/blob/master/query_update.go#L257).

jamesjulich:
Unless generating invalid SQL is intended (and any errors are intended to be caught at query execution, not query generation), then there are invalid test cases.

```
// internal\dbtest\query_test.go
// Line 191
func(db *bun.DB) schema.QueryAppender {
			return db.NewUpdate().Model(new(Model)).WherePK()
		},
```

This test query generates invalid SQL, and your test cases are checking against it.

hohobilly:
Also looking into the same issue.
Please correct me if I'm wrong, I wonder if // Line 191 `db.NewUpdate().Model(new(Model)).WherePK()` is valid because without `OmitZero()` it should imply for setting zero value for the empty fields. 
Instead, there is another test case in Line 571 which seems to fit this invalid generation issue.
```
  func(db *bun.DB) schema.QueryAppender {
    return db.NewUpdate().Model(&Model{ID: 42}).OmitZero().WherePK()
  },

> UPDATE `models` AS `model` SET WHERE (`model`.`id` = 42)
```

Issue #684, opened 2022-10-08
Column filtering with OmitZero causes invalid update statement


Karimerto:
I am trying to limit the number of columns that can be updated via an incoming REST request, but my idea is not working. I have the following (dummy) model:

    type Mytable struct {
        Id int    `json:"id" bun:",pk,autoincrement"`
        A  string `json:"a"`
        B  int    `json:"b"`
        C  string `json:"c"`
    }

And I would like that, for example, only columns A and B can be updated, and only if they actually contain something. So to that end I tried the following:

	var mytable Mytable
	// Parse from json, excluded for brevity
	if _, err := db.NewUpdate().
		Model(&mytable).
		Column("a", "b").
		OmitZero().
		WherePK().
		Returning("*").
		Exec(ctx); err != nil {
		return err
	}
	// Return jsonified response

When called with a proper request, for instance (in Python):

	requests.put("localhost:8080/api/mytable/123", json={"a": "some string"})

The result is fine, as expected. But when called with "invalid" request, the server fails with an internal error since the SQL generated is broken.

	requests.put("localhost:8080/api/mytable/123", json={"c": "this cannot be updated"})

The resulting SQL is something like the following:

	UPDATE "mytable" SET  WHERE ("mytable"."id" = 1) RETURNING *

As can be seen, the `SET` clause is empty, and naturally this is invalid SQL.

Is there a "proper" way this can be done, either with Bun itself, or Bunrouter? Or should I just add some filtering before running the update command?

jamesjulich:
Bumping since I am also experiencing this. Can confirm that I have reproduced a similar error.

If you ask bun to create an update statement using a model with all nil fields, then call OmitZero, you will end up with an invalid query as seen here.

```
db.NewUpdate().
	Where("? = ?", bun.Ident(identifier), c.Params(identifier)).
	OmitZero().
	Model(&myEmptyStruct{}).
	String();
```

will produce SQL like this: ```UPDATE "test" AS "test_struct" SET  WHERE ("id" = '1')```

PR #871, opened 2023-07-23
Add validation in generating empty SET clause for StructModel

hohobilly:
Attempt to fix the issues #858 & #684, where invalid `SET WHERE` statement is generated.
This only includes the check with struct model but not map. I can include map model if needed. 🙏 

Trying from a slightly different approach than the one suggested in the issue.

The test case mentioned in the issue: `db.NewUpdate().Model(new(Model)).WherePK()` should be a valid expression with `SET 'str' = ''` as `OmitZero()` is not set. It should set the field to ZeroValue.

Instead, test data snapshot `87` = `db.NewUpdate().Model(&Model{ID: 42}).OmitZero().WherePK()` should be the one invalid because no update field is provided while OmitZero is set.

Adding serval test cases as well.


vmihailenco:
Thanks. Ideally we should be able to skip such queries, but returning an error is a good first step too.

PR #887, opened 2023-08-23
feat: export there are no migrations error

greeflas:
Hi 👋
I have implemented a CLI command to run migrations. In new projects, I might not have migrations from the start, and in such cases, my command fails.
I want to address this issue by exporting the validation error to be able to check it:
```go
_, err := migrator.Migrate(ctx)

if errors.Is(err, migrator.ErrThereAreNoMigrations) {
	return nil
}
```

### DIFF
--- a/migrate/migrator.go
+++ b/migrate/migrator.go
@@ -12,6 +12,8 @@ import (
 	"github.com/uptrace/bun"
 )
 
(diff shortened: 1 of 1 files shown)


vmihailenco:
Sorry, but this looks like a minor internal detail that will likely change in future. I don't think this should not be part of the API.

PR #895, opened 2023-09-13
fix:  model not complete in beforeUpdate

qiruos:


vmihailenco:
The description is lacking and it looks the build is broken. Feel free to reopen with more details.

PR #897, opened 2023-09-15
fix: high memory usage

calorie:
v1.1.15 consumes a large amount of memory. Because https://github.com/uptrace/bun/pull/850 adds a lot of fields to a table.


vmihailenco:
I've just reverted that PR.

Issue #899, opened 2023-09-18
Allow dbfixture.New to Accept the IDB Interface


hirasawayuki:
## Description
Currently, the dbfixture.New function only accepts a pointer to bun.DB. However, both bun.DB and bun.TX implement the IDB interface. By allowing dbfixture.New to accept the IDB interface, it would provide more flexibility, especially during testing.

## Motivation
By allowing dbfixture.New to accept IDB, we can easily roll back transactions during testing. This would eliminate the need to manually delete test data after the tests have run, ensuring a clean database state and more efficient testing.

**Expected Usage**
Here's how it could be utilized with the suggest change:

```go
func ExampleTest(t *testing.T) {
	bdb := bun.NewDB(db, mysqldialect.New())
	tx, err := bdb.BeginTx(ctx, &sql.TxOptions{})
	if err != nil {
		t.Fatal(err.Error())
	}
	t.Cleanup(func() { _ = tx.Rollback() })

	fixture := dbfixture.New(tx)
	// ...
}
```

By making this change, we can leverage the power of transactions during testing, ensuring that our tests are both efficient and clean.

vmihailenco:
@hirasawayuki could you send a PR?

PR #900, opened 2023-09-18
feat: `dbfixture.New` to accept IDB interface

hirasawayuki:
This PR addresses the limitation where the dbfixture.New function only accepts a pointer to `bun.DB`. Both `bun.DB` and `bun.TX` implement the `IDB` interface. By allowing `dbfixture.New` to accept the `IDB` interface, we provide more flexibility, especially during testing.

close https://github.com/uptrace/bun/issues/899


vmihailenco:
Thanks!

Issue #903, opened 2023-09-22
Proposal: Add slog QueryHook for Bun logging


hirasawayuki:
## Description
This issue proposes the addition of a new package, tentatively named bunslog, to provide enhanced logging functionalities for the Bun using slog.

Bun currently supports various logging solutions as mentioned in the [debugging guide](https://bun.uptrace.dev/guide/debugging.html). However, there is no support for slog, a structured logging package. Notably, slog has been included in the Go standard library starting from version 1.21, which might a increase in demand.

PR #904, opened 2023-09-22
feat: bunslog.QueryHook for Bun logging using `log/slog`

hirasawayuki:
## Description
This PR implements a new package called bunslog to support enhanced logging functionalities for the Bun using `log/slog`.

ref #903 

## Features
- Supports setting a `*slog.Logger` instance or uses the global logger if not set.
- Logs general SQL queries with configurable log levels.
- Logs slow SQL queries based on a configurable duration threshold.
- Logs SQL queries that result in errors, for easier debugging.
- Allows for custom log formatting.

## Usage
Simply create a new QueryHook and add it to existing `*db.Bun` database instance.

```go
logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
slog.SetDefault(logger)

db := bun.NewDB(sqldb, dialect)
hook := bunslog.NewQueryHook(
    bunslog.WithQueryLogLevel(slog.LevelDebug),
    bunslog.WithSlowQueryLogLevel(slog.LevelWarn),
    bunslog.WithErrorQueryLogLevel(slog.LevelError),
    bunslog.WithSlowQueryThreshold(3 * time.Second),
)
db.AddQueryHook(hook)
```


extra/bunslog/bunslog.go
```diff
@@ -0,0 +1,113 @@
...
+func (h *QueryHook) BeforeQuery(ctx context.Context, event *bun.QueryEvent) context.Context {
+	return ctx
+}
+
+// AfterQuery is called after a query is executed.
+// It logs the query based on its duration and whether it resulted in an error.
+func (h *QueryHook) AfterQuery(ctx context.Context, event *bun.QueryEvent) {
+	duration := h.now().Sub(event.StartTime)
+	if event.Err != nil && !errors.Is(event.Err, sql.ErrNoRows) {
+		msg := fmt.Sprintf("[bunslog] error query: %s", event.Query)
```
vmihailenco, line 96: Can we use structured logging here instead? I.e. `slog.String("query", event.Query)`

vmihailenco:
@hirasawayuki Looks good, thanks. Could you add a `go.mod` file so we don't require `log/slog` is only available from the Go 1.21.

extra/bunslog/bunslog.go
hirasawayuki: @vmihailenco 
Yes, it's possible to use structured logging with `slog.String("query", event.Query)`. Would you like to omit the default `msg` field in that case? 

Regarding the expected log format, are you looking for something like this?
```json
{"time":"xxxx-xx-xxTxx:xx:xx+xx:xx","level":"DEBUG", "query":" SELECT `user`.`id`, `user`.`name`, `user`.`email` FROM `users`", "operation":"SELECT","duration":"3s"}
```
Or would you prefer this format?
```json
{"time":"xxxx-xx-xxTxx:xx:xx+xx:xx","level":"DEBUG", "msg":"[bunslog] query", "query":" SELECT `user`.`id`, `user`.`name`, `user`.`email` FROM `users`", "operation":"SELECT","duration":"3s"}
```

vmihailenco: >Would you like to omit the default msg field in that case?

I think so, yes.

>Adding an additional field, such as status is_slow, could help make this distinction clear.

For errors we should just add `slog.Any("error", event.Err)`. For slow queries I would not do anything special.

extra/bunslog/bunslog.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if h.slowQueryThreshold > 0 && h.slowQueryThreshold <= duration {
+		msg := fmt.Sprintf("[bunslog] slow query: %s", event.Query)
+		h.slogFns[h.slowQueryLogLevel](ctx, msg, slog.String("operation", event.Operation()), slog.String("duration", duration.String()))
+		return
+	}
+
+	msg := fmt.Sprintf("[bunslog] query: %s", event.Query)
+	h.slogFns[h.queryLogLevel](ctx, msg, slog.String("operation", event.Operation()), slog.String("duration", duration.String()))
```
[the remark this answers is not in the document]
hirasawayuki: Thank you for the review! I've fixed to use the `LogAttrs` method.

https://github.com/hirasawayuki/bun/blob/433ad6c04c293a570f3693fe055f2f45c7956964/extra/bunslog/bunslog.go#L123-L129

extra/bunslog/bunslog.go
hirasawayuki: I've omitted `msg` field and have fixed to allow for custom formats to be passed in as an option.

https://github.com/hirasawayuki/bun/blob/433ad6c04c293a570f3693fe055f2f45c7956964/extra/bunslog/bunslog.go#L93-L98

hirasawayuki:
@vmihailenco 
I've made some modifications to the implementation:
- Updated to allow for passing an instance of `slog.Logger` as an option, in addition to using the default logger.
- Added the ability to set a custom format through an option.
I would appreciate another review! Thank you!

(It appears that the build is failing for Go 1.19 and Go 1.20 due to the import of log/slog. Would it be a good idea to also modify the Makefile to avoid loading it during the build?)

vmihailenco:
>Would it be a good idea to also modify the Makefile to avoid loading it during the build?)

Let's just bump Go in tests to v1.21.

vmihailenco (approved):

vmihailenco:
Thanks!

PR #916, opened 2023-10-06
add CreateTxSQLMigrations function

codeliger:
This PR enables the `CreateSQLMigrations` function to create transactional sql migration files (`.tx.sql`)


migrate/migrator.go
```diff
@@ -268,29 +268,42 @@ func (m *Migrator) CreateGoMigration(
 }
 
 // CreateSQLMigrations creates an up and down SQL migration files.
-func (m *Migrator) CreateSQLMigrations(ctx context.Context, name string) ([]*MigrationFile, error) {
+func (m *Migrator) CreateSQLMigrations(ctx context.Context, name string, transactional bool) ([]*MigrationFile, error) {
```
vmihailenco, line 271: Could you please leave this method as is and instead add `CreateTxSQLMigrations`? This way we don't break existing API.

codeliger: Undid the changes and made the function CreateTxSQLMigrations

vmihailenco:
Thank you :+1:

Issue #801, opened 2023-03-14
Limit 0 returns all row set instead of empty set


KaitoHH:
The following lines
https://github.com/uptrace/bun/blob/7cd855e64a02459579c8dd2b7bedfdc00df89cf3/query_select.go#L634-L637

causes bun omits to append `Limit 0` when selecting rows, resulting in returning all of the rows, which is not consistent with the expected empty set

wikylyu:
I think `limit 0` is meaningless. If you want to `limit 0`, why not just return [] without selecting database.

Issue #811, opened 2023-04-04
Erro dialect mssql witch NEXT in the FETCH


chapzin:
in import i use:
```
_ "github.com/denisenkom/go-mssqldb"
"github.com/uptrace/bun"
"github.com/uptrace/bun/dialect/mssqldialect"
```
function i use:
```
func (i *Inventario) GetFistThreeItens(db *bun.DB, cnpj string, ano int) ([]models.Inventario, error) {
	var itens []models.Inventario
	err := db.NewSelect().Model(&itens).Where("cnpj = ? and ano = ?", cnpj, ano).Limit(3).Scan(context.Background())
	if err != nil {
		return nil, err
	}
	return itens, nil
}
```
recive erro return: mssql.Error: mssql: Invalid usage of the option NEXT in the FETCH statement

my microsoft sql server 2017

chapzin:
problem fix if use Order before Limit
look now
```
func (i *Inventario) GetFistThreeItens(db *bun.DB, cnpj string, ano int) ([]models.Inventario, error) {
	var itens []models.Inventario
	err := db.NewSelect().Model(&itens).Where("cnpj = ? and ano = ?", cnpj, ano).Order("id").Limit(3).Scan(context.Background())
	if err != nil {
		return nil, err
	}
	return itens, nil
}
```

PR #918, opened 2023-10-19
fix: allow LIMIT 0 for SELECT queries

ygabuev:
This PR enables the correct queries of form `SELECT ... LIMIT 0`.
Before that, the limit-clause wasn't applied to the query.

Closes https://github.com/uptrace/bun/issues/801


ygabuev:
The new tests fail for `mssql`, I've checked and they also fail for current master. Given that there were no tests for the `LIMIT` clause, I have a felling that the implementation for `LIMIT` and/or `OFFSET` is broken for `mssql`.

query_select.go
```diff
@@ -31,7 +31,7 @@ type SelectQuery struct {
 	group      []schema.QueryWithArgs
 	having     []schema.QueryWithArgs
 	order      []schema.QueryWithArgs
-	limit      int32
+	limit      *int32
```
vmihailenco, line 34: let's use `sql.NullInt32` here

vmihailenco:
@ygabuev this looks good, but I can't merge a PR that breaks the build....

CL-Jeremy:
> The new tests fail for `mssql`, I've checked and they also fail for current master. Given that there were no tests for the `LIMIT` clause, I have a felling that the implementation for `LIMIT` and/or `OFFSET` is broken for `mssql`.
> 
> If this is true, it might be a better idea to fix those in a separate PR.
> 
> UPD: this ticket seems related #811

Hi, I just implemented a fix for that problem in #1009. Could you please test if this workaround fixes your problem?

PR #926, opened 2023-11-01
AutoMigrate for Postgres dialect

bevzzz:
This PR is my second attempt at implementing #456 :)

This time around I decided not to meddle with `ALTER TABLE` query and focus on the functionality we want to bring -- AutoMigrator. I took into account our discussions in  #726 and left it up to the dialects to implement "alter table" operations, keeping it internal. My thinking was that, if auto migration is done well, the users wouldn't need to work with ALTER TABLE all that much.

AutoMigrator exposes a simple and minimal API: `CreateSQLMigrations(ctx)` and `Migrate(ctx, options)`. These are similar to Migrator except that they do not require writing any SQL/Go migrations manually.

What AutoMigrator can do:
- Create, drop, and rename tables
- Create, drop, and rename column
- Modify `DEFAULT` value
- Modify `NOT NULL` and `UNIQUE` constraints
- Modify primary key and foreign key constraints
- Change column data type
- Add or remove identity
- Create SQL migration files with support for transactional migrations (`.tx.up.sql`/`.tx.down.sql`)

What it doesn't do:
AutoMigrator performs _schema migrations_ and not _data migrations_. For example, when changing a primary key on the table, related foreign keys will not be updated -- they will continue referencing the old PK; to "re-link" them would require updating the referencing values themselves, which is beyond AutoMigrator's scope. 

Presently, only `pgdialect` supports database inspection and migration, for which it implements two additional interfaces. AutoMigrator will return an informative error if used with a dialect that doesn't support this feature.

Other changes:
- `test.sh` forwards additional arguments to `go test` 
(for example, now possible to run 1 test with `./test.sh -run=TestAutoMigrator_Run`)
- added `Schema string` field to differentiate tables in different schemas. 
- each Dialect now reports the name of the default schema
- had to update commit message in 95c4a8ebd634e1e99114727a7b157eeeb9297ee9 because it was upsetting the linter
___

-  [x] In-code documentation
-  [ ] Documentation on [the website](https://bun.uptrace.dev)
-  [ ] Example CLI to generate and run auto-migrations


bevzzz:
I had to a lot of changes to the existing unit tests (~5-7 files) because I noticed that many of them weren't properly cleaning up the database, causing side-effects in the auto-migrate tests.  
All of them were done in a [single commit](https://github.com/uptrace/bun/pull/926/commits/251f4b26859eff5dac40d6ebb1f56f127124f72c) and I could open a different PR for them to be merged separately.

migrate/sqlschema/state.go
```diff
@@ -0,0 +1,386 @@
...
+
+import (
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/uptrace/bun/schema"
+)
+
+type State struct {
```
vmihailenco, line 11:
```go
type State interface {
    Tables() []Table
}

type Table interface {
   Columns() []Column
}

table.(*pgdialect.Table)
table.(*mysqldialect.Table)
```

migrate/operations.go
```diff
@@ -0,0 +1,377 @@
...
+)
+
+// Operation encapsulates the request to change a database definition
+// and knowns which operation can revert it.
+type Operation interface {
+	GetReverse() Operation
+}
+
+// CreateTable
+type CreateTable struct {
```
vmihailenco, line 15: CreateTable -> CreateTableOp

bevzzz:
Looks like we're now creating those up and down files!

image

migrate/sqlschema/state.go
bevzzz: After some experimenting with what these interfaces might be I arrived at a slightly different solution. 
Here's a short example:

```go
// sqlschema
type Table struct {
  Name  string
  Columns  []Column

  // Additional stores opaque metadata specific to each dialect
  Additional interface{}
}

// pgdialect
type Table struct {
  InheritsFrom  sqlschema.Table // Postgres supports table-inheritance
} 

table := Table{..., InheritsFrom: parentTable}

// elsewhere
pgtable := table.Additional.(*pgdialect.Table)
```

This achieves the same flexibility -- different dialects or even implementations of `Inspector` may extend the table with additional information and type-cast it later. That way we wouldn't need to define interfaces for what is essentially a DTO.

WDYT?

vmihailenco: >different dialects or even implementations of Inspector may extend the table with additional information and type-cast it later.

It works indeed but looks awkward to me.

This looks much more straightforward and achieves the same:

```go
// sqlschema

type BaseTable struct {
  Name    string
  Columns []Column
}

func (t *BaseTable) GetName() { return t.Name }
func (t *BaseTable) GetColumns() { return t.Columns }

type Table interface {
  GetName() string
  GetColumns() []Column
}

// pgdialect

type Table struct {
  sqlschema.BaseTable

  InheritsFrom  *Table // Postgres supports table-inheritance
} 

// usage

var table sqlschema.Table
table = &pgschema.Table{
  BaseTable: sqlschema.BaseTable{Name: ..., Columns: ...},
  InheritsFrom: parentTable,
}}

// elsewhere
pgtable := table.(*pgdialect.Table)
pgtable.InheritsFrom
```

But we can definitely start with you design if you insist.

dialect/pgdialect/alter_table.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return b, nil
+}
+
+func (m *migrator) addColumn(fmter schema.Formatter, b []byte, add *migrate.AddColumnOp) (_ []byte, err error) {
+	b = append(b, "ADD COLUMN "...)
+	b = fmter.AppendName(b, add.Column)
+	b = append(b, " "...)
+
+	colDef, ok := add.ColDef.(sqlschema.ColumnDefinition)
```
vmihailenco, line 94: ColDef.(sqlschema.ColumnDefinition) looks strange and repetitive (масло маслянное).

And names like `Column` and `ColumnDefinition` look like they have distinct meaning while they are just inerface and an implementation.

dialect/pgdialect/alter_table.go
```diff
@@ -0,0 +1,248 @@
...
+	if !ok {
+		return nil, fmt.Errorf("column %q does not implement sqlschema.ColumnDefinition, got %T", add.ColDef.GetName(), add.ColDef)
```
vmihailenco, line 96: ColumnDefinition is a struct so it is actually ColumnDefinition which implements Column. I will change these names...

vmihailenco:
@bevzzz few things that I've noticed:

- You've added schema.FQN to support `schema.table` names. Do we really need it? It complicates things a lot and I don't think that we should support more than a single schema, e.g. do we really plan to support moving tables between schemas? Is that even possible?

  I suggest we drop `schema.FQN` and ignore the schema name althogether.

- In every interface you have `Tables []Table`, but `Tables map[string]Table`. That will cause some issues, because maps don't preserve the order. I suggested we use ordered maps in all such places, for example, https://github.com/wk8/go-ordered-map

So we will have:

```go
type Table interface {
  GetTables() *orderedmap.Map[string, Table]
}
```

- I also added some clarification comment to BaseTable, BaseColumn etc so they must not be used outside dialects

```go
// BaseTable is a base table definition.
//
// Dialects and only dialects can use it to implement the Table interface.
// Other packages must use the Table interface.
```

WDYT?

bevzzz:
> do we really plan to support moving tables between schemas? Is that even possible?

No, I don't think moving schemas between tables is possible. Rather, I wanted to make sure that the tables in different schemas aren't shadowing each other names, i.e. I want to be able to rename `books.authors -> writers` without renaming `public.authors`  or `paintings.authors`.

Most databases I've worked with had several schemas, so I thought this may make automigrations more resilient to name clashes.

> That will cause some issues, because maps don't preserve the order. I suggested we use ordered maps

Sure, I would even argue that `GetTables` may  return `[]Table` and have  `detector` use `orderedmap.Map` internally, where it's most needed.  This way dialects needn't bring `orderedmap` as a dependency.

Other than the different order in which Operations may be applied, is there any reason a lack of order in `map` may cause issues? I'm asking because the dependencies between operations are always resolved during migration planning and the order in which tables/columns are supplied does not break that.

dialect/pgdialect/alter_table.go
bevzzz: Thank you

vmihailenco:
> is there any reason a lack of order in map may cause issues?

I assume that query generation won't be stable if 2 columns/tables are changed at once.

I also see that the tests rely on having a `map[string]Table/Column`. Also I had a similar task before and orderedmap worked well.

vmihailenco:
@bevzzz I am going to introduce `orderedmap` myself and leave FQN removal to you. Hopefully that's okay with you.

Regarding schemas, I think that we might need to accept a schema name via configuration and then only work with that schema. But that can be added later.

vmihailenco:
@bevzzz я все :)

Тесты я поправил но билд почему-то не проходит c непонятной ошибкой. Предлагаю пока это игнорировать.

bevzzz:
Short update: we've decided to drop multi-schema support, so the user would instead configure AutoMigrator to work with a particular schema by passing `migrate.WithSchemaName("my_schema")` option.
By default, AutoMigrator will continue to use the dialect's default schema, e.g. `public` for Postgres.

vmihailenco:
Looks like we can now remove `(Dialect) DefaultSchema()`. Overall this looks pretty good.

migrate/sqlschema/inspector.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+
+	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/schema"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
+)
+
+type InspectorDialect interface {
+	schema.Dialect
+	Inspector(db *bun.DB, excludeTables ...string) Inspector
```
vmihailenco, line 16: `excludeTables` should be an option so we can extend this later.

`Inspector(db *bun.DB, options ...sqlschema.InspectorOption) Inspector`

vmihailenco, line 16: `excludeTables` should be an option so we can extend this later.

`Inspector(db *bun.DB, options ...sqlschema.InspectorOption) Inspector`

migrate/sqlschema/inspector.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// EquivalentType returns true if col1 and co2 SQL types are equivalent,
+	// i.e. they might use dialect-specifc type aliases (SERIAL ~ SMALLINT)
+	// or specify the same VARCHAR length differently (VARCHAR(255) ~ VARCHAR).
+	EquivalentType(Column, Column) bool
+}
+
+// Inspector reads schema state.
+type Inspector interface {
+	Inspect(ctx context.Context, schemaName string) (Database, error)
```
vmihailenco, line 26: I'd say schema name should be passed when creation an inspector:

```
dialect.Inspector(db, sqlschema.WithSchemaName(...), sqlschema.ExcludeTables(...))
```

Maybe event rename `Inspector` to `NewInspector` for clarity.

vmihailenco:
This looks pretty good. @bevzzz you should have necessary permissions to merge this so feel free to do it.

vmihailenco (approved):

migrate/sqlschema/inspector.go
bevzzz: Addressed in #1058

migrate/sqlschema/inspector.go
```diff
@@ -0,0 +1,200 @@
...
+)
+
+type InspectorDialect interface {
+	schema.Dialect
+	Inspector(db *bun.DB, excludeTables ...string) Inspector
+
+	// EquivalentType returns true if col1 and co2 SQL types are equivalent,
+	// i.e. they might use dialect-specifc type aliases (SERIAL ~ SMALLINT)
+	// or specify the same VARCHAR length differently (VARCHAR(255) ~ VARCHAR).
+	EquivalentType(Column, Column) bool
```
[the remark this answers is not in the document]
bevzzz: Addressed in #1058

migrate/sqlschema/inspector.go
bevzzz: Addressed in #1058

migrate/sqlschema/inspector.go
bevzzz: This is a duplicate of the suggestion above.

bgrant0607:
Hi. Thanks for this change. It looks like it will be useful. My use case is that I just wanted to generate and write out the migrations automatically, but not execute them at the same time.

One piece of feedback regarding ordered.Map as a dependency: ordered map is very small. One small, simple file. The choice to use an internal type in exported fields of exported structs made this code unusable as is, and made extending and building upon it really really messy. I revisited this PR to see why that decision had been made when so many other types are fully exported.

One other tiny issue that had cascading consequences: Anyone who uses WithForeignKeys in their tables instead of or in addition to relation tags will have those foreign keys dropped. Adding a WithForeignKeys AutoMigratorOption is a small change (a few lines of code) that would address that. I may submit a PR for that at some point.

Also, it would be handy for migrate.comment to be an exported type. As it is, I had to use reflection to detect them.

bevzzz:
hi @bgrant0607
thank you for taking the time to share your thoughts! you're right, that's something we've overlooked in the original implementation. I'll continue the discussion in the issue you've opened 👍 
 
> migrate.comment to be an exported type

Fair point, no harm in exporting it, given all other operation types are exposed as well.
Might need to think of a better name for the type; I'll see to it.

> My use case is that I just wanted to generate and write out the migrations

Can your use case be served by `CreateSQLMigrations()`? It will generate migration files, but not execute them.  
If not, would it be useful to have AutoMigrator write SQL to a custom Writer? `*`

> The choice to use an internal type in exported fields of exported structs made this code unusable as is

Good point, I am going to look into that. Perhaps we could somehow push `orderedmap` to the inspectors' internals and export a slice instead.

---

`*` - Could you describe your current application in a bit more detail? You've mentioned that you're catching `migration.comment` with reflection, which has me a bit puzzled, because `changeset` is an unexported type, produced and iterated-over by AutoMigrator internally.

bgrant0607:
@bevzzz I ran into multiple problems with the automigrate code and I was trying to build on it to work out solutions without forking all of bun.

I reported two issues: #1159 and #1160.

Another issue I had was that we have some tables with generated names that are managed separately, so I wanted to exclude tables based on a prefix or pattern.

And I needed to be able to set the directory and the output filenames. The options weren't plumbed through CreateSQLMigrations and/or exported through options in a way I could use. 

I also had some issues with serializing multiple migration queries, like lack of semicolons and newlines IIRC, so I had to fork that part also to add the 2 characters. Maybe I was missing something from the pgdialect implementation.

My diff of bun/migrate ended up being 1777 lines even though my changes were very minor.

bgrant0607:
My patched version of WriteTo:
```
func (c *changeset) WriteTo(w io.Writer, m bunsqlschema.Migrator) error {
        var err error

        b := MakeQueryBytes()
        for _, op := range c.operations {
                if isComment := reflect.TypeOf(op).String() == "*migrate.comment"; isComment {
                        // Just skip it so that we don't need to try to extract the value from a private type
                        // b = append(b, "/*\n"...)
                        // b = append(b, ...)
                        // b = append(b, "\n*/"...)
                        continue
                }

                b, err = m.AppendSQL(b, op)
                if err != nil {
                        return fmt.Errorf("write changeset: %w", err)
                }
                b = append(b, ";\n"...)
        }
        if _, err := w.Write(b); err != nil {
                return fmt.Errorf("write changeset: %w", err)
        }
        return nil
}
```

bevzzz:
> ```
> if isComment := reflect.TypeOf(op).String() == "*migrate.comment"; isComment {
>   // Just skip it so that we don't need to try to extract the value from a private type
>   // b = append(b, "/*\n"...)
>   // b = append(b, ...)
>   // b = append(b, "\n*/"...)
>   continue
> }
> ```

I recall you also mentioned this, that working with an unexpected `comment` type was unnecessarily cumbersome. We've since [changed it](https://github.com/uptrace/bun/pull/1165) to an exported `migrate.Unimplemented`.

> I wanted to exclude tables based on a prefix or pattern

That's a good idea. We can add `WithExcludeTablesLike()` option that would accept the same patten syntax as SQL's `LIKE` clause. E.g., `"generated_tbl_%"` to exclude all "generated_tbl_" tables.

> I also had some issues with serializing multiple migration queries, like lack of semicolons and newlines IIRC, so I had to fork that part also to add the 2 characters. Maybe I was missing something from the pgdialect implementation.

If you have the time to do it and the issue persists, would you care to post an MRE for me to look into? We do not any tests for this specifically, but it may be worth adding.

> And I needed to be able to set the directory and the output filenames.

There's `WithMigrationsDirectoryAuto()` option:

https://github.com/uptrace/bun/blob/04da1b6d1c6371a471d34822bf4d3bc429d57b1e/migrate/auto.go#L79

which is the AutoMigrator's double for `WithMigrationsDirectory()` ⬇️ 

https://github.com/uptrace/bun/blob/04da1b6d1c6371a471d34822bf4d3bc429d57b1e/migrate/migrations.go#L16

I've added a TODO about providing another option to control filenames too 👍 .

> My diff of bun/migrate ended up being 1777 lines even though my changes were very minor.

I've started #1166 to track the progress on the improvements to AutoMigrator. If you feel like any of the other changes you've made will make a useful addition, please share your ideas there! Once again, appreciate the feedback you've provided so far, it's been very useful.

PR #927, opened 2023-11-03
Cleanup test databases to avoid side-effects

bevzzz:
While working on #926 I noticed that many tests did not properly tear down the database schema they create, which results in side-effects to other tests.

Existing tests were not affected,  but the ones I wrote for the other PR were, as they rely on having a clean schema.
In general, I think leaving no side-effects is a good unit test practice.


vmihailenco:
Thanks!

Issue #929, opened 2023-11-11
 db.NewCreateTable with `bun:",pk,autoincrement"` did not create table   sqlite_sequence


tablecell:
```
package main

import(
"fmt"
"context"
    "database/sql"
     "github.com/uptrace/bun"
  "github.com/uptrace/bun/driver/sqliteshim"
"github.com/uptrace/bun/dialect/sqlitedialect"
)

type User struct {
 bun.BaseModel `bun:"table:users,alias:u"`
 ID	 int64  `bun:",pk,autoincrement"`
 Name string


}
func main(){
ctx:=context.Background()
sqldb, err := sql.Open(sqliteshim.ShimName, "test.db")  // "file::memory:?cache=shared"
if err != nil {
	panic(err)
}
db := bun.NewDB(sqldb, sqlitedialect.New())
 fmt.Println(db)
// user := &User{Name: "admin"}

ret, err := db.NewCreateTable().Model((*User)(nil)).Exec(ctx)
fmt.Println(ret,err )
}

```
generated create table  sql is:
```
CREATE TABLE "users" ("id" INTEGER NOT NULL, "name" VARCHAR, PRIMARY KEY ("id"));
```
there is no " autoincrement "

bevzzz:
As far as I can tell, the reason the sequence was not created is that in SQLite you get unique autoincremented ID column (called `ROWID`) for any table that has `INTEGER PRIMARY KEY` column in it. 

References:   
- https://www.sqlite.org/lang_createtable.html#rowid (Section **5. ROWID**) 
- https://stackoverflow.com/a/7906029/14726116

That being said, there _is_ a [dedicated `AUTOINCREMENT` keyword](https://www.sqlite.org/autoinc.html) in SQLite, which ensures that the next ROWID is _is at least one larger than the largest ROWID that has ever before existed in that same table_.   

I imagine that the current lack of support for it in `bun` was due to the default mechanism fulfiling most of the requirements you'd have for an autoincremented column.  You're right in that it'd make sense if `bun:",autoincrement"` created the sequence instead of silently falling back to `ROWID`.  

Looks like a small change, I'll try to come up with a PR for this.

Issue #688, opened 2022-10-09
Postgre `identity` lack of `GENERATED ALWAYS` support


necipallef:
[Here](https://github.com/uptrace/bun/blob/28915f89b0e864ee7c43782fe3234f8d65831e00/query_table_create.go#L173) it seems like the `identity` tag results in `GENERATED BY DEFAULT` only. However, Postgre has an option `GENERATED ALWAYS` as seen [in this link](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-identity-column/):
`The 'GENERATED ALWAYS' instructs PostgreSQL to always generate a value for the identity column. If you attempt to insert (or update) values into the 'GENERATED ALWAYS AS IDENTITY' column, PostgreSQL will issue an error.`

Also, I could not find any documentation for the `identity` tag. I was able to find the `identity` tag ever exists by searching through the issues. Is it me or is there a lack of information on the `identity` tag?

GauteHaugen:
I can see that this issue has been here for some time now. Are there any updates or plans to work on this? In addition to the lack of GENERATED ALWAYS keywords, there are also missing the optional sequence_options that go with the identity column.

This definition is taken from PostgreSQL documentation:
```GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY [ ( sequence_options ) ]```

bevzzz:
@vmihailenco WDYT about extending the tag to allow `identity:always`, defaulting to `BY DEFAULT` if the tag has no options?

PR #936, opened 2023-11-29
Support AUTOINCREMENT directive in SQLite

bevzzz:
At the moment `sqlite` dialect ignores `bun:",autoincrement"` tag, relying on the default `ROWID` behaviour instead.  
However, [`AUTOINCREMENT` keyword](https://www.sqlite.org/autoinc.html) is a separate instruction in SQLite that creates a dedicated sequence and changes how the unique IDs for the table are generated.  

Related issue: #929 

This PR adds support for `AUTOINCREMENT` keyword for fields which have the `autoincrement` tag. See the docs for [`sqlitedialect.Dialect.AppendSequence`](https://github.com/uptrace/bun/compare/master...bevzzz:bun:fix/929-sqlite-autoincrement?expand=1#diff-a997dc82d234a0a796e5c3e5473862b9e9fcc2677353aec003c619c632b42d4cR107) for more details.  

I also noticed that each dialect has different keywords for creating generated columns, so I also pushed this logic to the dialects' via new `AppendSequence` method.


dialect/sqlitedialect/dialect.go
```diff
@@ -91,6 +92,25 @@ func (d *Dialect) DefaultVarcharLen() int {
...
+// AppendSequence adds a primary key constraint as a *side-effect*. Callers should expect it to avoid building invalid SQL.
+// SQLite also [does not support] AUTOINCREMENT column in composite primary keys.
+//
+// [documentation]: https://www.sqlite.org/autoinc.html
+// [does not support]: https://stackoverflow.com/a/6793274/14726116
+func (d *Dialect) AppendSequence(b []byte, table *schema.Table, field *schema.Field) []byte {
+	if field.IsPK && len(table.PKs) == 1 && field.CreateTableSQLType == sqltype.Integer {
+		b = append(b, " PRIMARY KEY AUTOINCREMENT"...)
+	}
+	return b
```
vmihailenco, line 111: Could you refactor this method to always append `PRIMARY KEY AUTOINCREMENT` and move the logic upwards to `query_table_create.go`?

query_table_create.go
```diff
@@ -309,7 +303,8 @@ func (q *CreateTableQuery) appendFKConstraints(
 }
 
 func (q *CreateTableQuery) appendPKConstraint(b []byte, pks []*schema.Field) []byte {
-	if len(pks) == 0 {
+	// Primary key constraint might've already been created together with AUTOINCREMENT directives.
+	if len(pks) == 0 || bytes.Contains(b, []byte("PRIMARY KEY")) {
```
vmihailenco, line 307: Please move this to the caller side so this method is called only if `...`.

Perhaps we could also make it more clear `bytes.Contains(b, []byte("PRIMARY KEY"))` applies to SQLite.

vmihailenco (approved):

vmihailenco:
@bevzzz looks good with few notes :+1:

query_table_create.go
bevzzz: Pushed the check to the caller and added more info in the comment

dialect/sqlitedialect/dialect.go
bevzzz: Have I understood your suggestion correctly?  

```go
// query_table_create.go

if (field.Identity && fmter.HasFeature(feature.GeneratedIdentity)) ||
	(field.AutoIncrement && (fmter.HasFeature(feature.AutoIncrement) || fmter.HasFeature(feature.Identity))) {
		
	if q.db.dialect.Name() == dialect.SQLite {
		if field.IsPK && len(q.table.PKs) == 1 && field.CreateTableSQLType == sqltype.Integer {
			b = q.db.dialect.AppendSequence(b, q.table, field)
		}
	} else {
		b = q.db.dialect.AppendSequence(b, q.table, field)
	}
}
```

If I may share some thoughts about the code above:

1. Surfacing that `if` condition in the `CreateTable` makes it easier to see when the method will actually do something.  
1. The limitations of `AUTOINCREMENT` keyword (SQLite-specific) seem like something that'd  be best described  in the dialect code.
1. Other callers of `AppendSequence()` (if there are any)  would need to know about it and do the check themselves.


WDYT?

vmihailenco: >Other callers of AppendSequence() (if there are any) would need to know about it and do the check themselves.

Well, the details are already leaking to other places, e.g. you have to craft `bytes.Contains` check to account for what is done in `AppendSequence`.

I would probably just remove `AppendSequence` since the abstraction it provides is not that useful and leaks to other places.

We can do it later though.

bevzzz: >  the abstraction it provides is not that useful

The use case I had in mind was the ability to add dialect-specific details, like switching between `GENERATED { BY DEFAULT | IDENTITY} AS SEQUENCE` in Postgres (see #688) or setting [`seed` and `increment`](https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property?view=sql-server-ver16#syntax) in MSSQL.

> the details are already leaking to other places, e.g. you have to craft bytes.Contains

You're right, I haven't thought about that before.  
Do you think it'd be an improvement if instead `AppendSequence()` returned a `bool` indicating whether "a PRIMARY KEY can still be defined in this query"? With that being clearly documented in the `Dialect` interface.

We'll need to check if we can still `q.appendPKConstraint()` after handling the sequence either way.

PR #939, opened 2023-12-11
fix: auto add alias for order column added via Column() method if the colu…

ilxqx:
Auto add alias for order column added via Column() method if the column in the table FieldMap


vmihailenco:
This requires better explanation of what you are trying to achieve and what are the alternatives. TBH I am not big fan of having queries the result of which depend on the previous state of `FieldMap`... It probably will be too brittle...

ilxqx:
If I have a SQL query with multiple table joins, when I specify to sort by `Order`, if it's the sorting field of the main table, I'll omit the table alias; if it's a sorting field from other joined tables, then I will directly specify the `alias.field_name`.

This semantic is consistent with the `Column` method and I think this better reflects its practical meaning for framework users.

vmihailenco:
@ilxqx okay, could you please update query_test.go with some examples that demonstrate your point?

PR #941, opened 2023-12-17
CreateTableQuery: Small cleanups for foreign key handling

bevzzz:
These are just some cosmetic changes I've stumbled upon while working on another ticket:
- swapped `HasOneRelation` and `BelongsToRelation` places, so that the type is set correctly. That did not affect the functionality, but could prevent some obscure bugs in the future.
- added a `References() bool` method to `Relation` to highlight the difference between the ones that need a `FOREIGN KEY` clause themselves and the ones that are referenced by other tables instead.
- idempotent `.WithForeignKeys()`


query_table_create.go
```diff
@@ -212,6 +201,10 @@ func (q *CreateTableQuery) AppendQuery(fmter schema.Formatter, b []byte) (_ []by
 
 	b = q.appendPKConstraint(b, q.table.PKs)
 	b = q.appendUniqueConstraints(fmter, b)
+
+	if q.autoFKs {
+		q.createRelationFKs()
```
vmihailenco, line 206: As is, the query will be broken if it is appended multiple times.

Instead, you should probably rewrite the `createRelationFKs` so you generate the query by appending to `b`.

bevzzz: Yep, you're totally right.  
Thanks for catching that, just pushed a test and a fix for it 🔧

PR #952, opened 2024-01-31
Allow overiding of `Warn` and `Deprecated` loggers

thomasvn:
Currently,  `internal.Warn.Printf()` and `internal.Deprecate.Printf()` have numerous calls but no way of overriding them with your own logger.


thomasvn:
@vmihailenco Any thoughts on this?

vmihailenco:
We probably want to start using Slog, but this will do for now. Thanks!

thomasvn:
Agree, thanks for the review!

PR #953, opened 2024-02-01
enable SNI

conradludgate:
Adding the ServerName config allows TLS to include the ServerNameIdentification (SNI) extension. We use this at Neon to determine which database endpoint to connect to: https://neon.tech/docs/connect/connection-errors#the-endpoint-id-is-not-specified

I have tested that this works for `sslmode=require`, but I need to still confirm that this doesn't break the insecure modes from being insecure

### DIFF
--- a/driver/pgdriver/config.go
+++ b/driver/pgdriver/config.go
@@ -269,6 +269,10 @@ func parseDSN(dsn string) ([]Option, error) {
 		case "require":
 			if sslRootCert == "" {
 				tlsConfig.InsecureSkipVerify = true
+				tlsConfig.ServerName = u.Host
+				if host, _, err := net.SplitHostPort(u.Host); err == nil {
+					tlsConfig.ServerName = host
+				}
 				break
 			}
 			// For backwards compatibility reasons, in the presence of `sslrootcert`,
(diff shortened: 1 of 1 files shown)


eu-erwin:
Hi @conradludgate, I currently face the same problem on neon too.
however it works with `sslmode=verify-full`. does this have some impact? I am not sure, what the different is.
but, at lease it works for now.

conradludgate:
sslmode=require does not verify that the certificate is valid, it only ensures that TLS is used. I would recommend verify-full of you can (neon supports this perfectly)

driver/pgdriver/config.go
```diff
@@ -283,6 +287,10 @@ func parseDSN(dsn string) ([]Option, error) {
 			// (verify chain, but skip server name).
 			// See https://github.com/golang/go/issues/21971 .
```
vmihailenco, line 288: Please update the comment and explain why this is required

conradludgate: The comment is still correct. `InsecureSkipVerify = true` will not verify either the certificate chain nor verify the server name. The `ServerName` field only adds the server name to the TLS ClientHello but doesn't impact the verification if insecure is set.

vmihailenco: Good know, thanks.

PR #959, opened 2024-02-12
fix: transactional migration file extension

colethienes:
The migrations wouldn't run unless formatted like `tx.up.sql`, which is the correct extension according to the docs: https://bun.uptrace.dev/guide/migrations.html#sql-based-migrations


vmihailenco:
:+1:

PR #969, opened 2024-04-01
feat: add ModelTableExpr to TruncateTableQuery

### LINKED ISSUE #960: TrunateTableQuery: missing ModelTableExpr() method
Hello, this is probably more of a question, but why does this query type doesn't feature the `ModelTableExpr` overriding method?

https://github.com/uptrace/bun/blob/master/query_table_truncate.go

JunNishimura:
This PR fixes #960.

I implemented a ModelTableExpr method in TruncateTableQuery as well, referring to CreateTableQuery, etc.

The [Document](https://bun.uptrace.dev/guide/query-truncate-table.html#api) stated that ModelTableExpr can be used, but it is not implemented actually, so I made this modification.

I would appreciate it if you check it out!

### DIFF
--- a/migrate/migrator.go
+++ b/migrate/migrator.go
@@ -362,7 +362,9 @@ func (m *Migrator) MarkUnapplied(ctx context.Context, migration *Migration) erro
 }
 
 func (m *Migrator) TruncateTable(ctx context.Context) error {
-	_, err := m.db.NewTruncateTable().TableExpr(m.table).Exec(ctx)
+	_, err := m.db.NewTruncateTable().
+		ModelTableExpr(m.table).
+		Exec(ctx)
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


migrate/migrator.go
```diff
@@ -362,7 +362,9 @@ func (m *Migrator) MarkUnapplied(ctx context.Context, migration *Migration) erro
 }
 
 func (m *Migrator) TruncateTable(ctx context.Context) error {
-	_, err := m.db.NewTruncateTable().TableExpr(m.table).Exec(ctx)
+	_, err := m.db.NewTruncateTable().
+		ModelTableExpr(m.table).
+		Exec(ctx)
```
vmihailenco, line 367: This does not look right, because there is no `Model` call. Could you revert it?

JunNishimura: @vmihailenco 
Thanks for the reply.

Referring to other implementations, I decided that it would be better to call a Model method than to revert.
I would appreciate it if you could check the following commit again. https://github.com/uptrace/bun/pull/969/commits/27193c377d854c591f07b640349d8de7f9bf5bb0

PR #972, opened 2024-04-02
Fix embedded field detection

adriansmares:
This PR fixes the situation in which an `embed` tag is combined with a set of `scanonly` fields. 

This issue seems to be caused by https://github.com/uptrace/bun/commit/9052fc4cdf3a1e06facae872d0f2c12d0b1e5bfb - see the review comments below.

cc @vmihailenco

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -207,7 +207,7 @@ func (t *Table) processFields(
 			}
 
 			subtable := newTable(t.dialect, fieldType, seen, canAddr)
-			for _, subfield := range subtable.Fields {
+			for _, subfield := range subtable.allFields {
 				embedded = append(embedded, embeddedField{
 					prefix:     prefix,
 					index:      sf.Index,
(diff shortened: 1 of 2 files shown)


schema/table.go
```diff
@@ -207,7 +207,7 @@ func (t *Table) processFields(
 			}
 
 			subtable := newTable(t.dialect, fieldType, seen, canAddr)
-			for _, subfield := range subtable.Fields {
+			for _, subfield := range subtable.allFields {
```
adriansmares, line 210: `newTable` does not put any fields in `subtable.Fields`, so this for loop is guaranteed to never iterate over something.

It is also the case that on L177 we also iterate over `subtable.allFields`, so maybe this was some kind of copy paste error ?

vmihailenco:
@adriansmares thanks for fixing this with a test :+1:

PR #974, opened 2024-04-04
Refactor unit test of query_test.go

JunNishimura:
Since it is difficult to find out which file in the testdata/snapshots directory is associated with the query execution results of each test case in query_test.go, I assign a number for each test case.

The contents of the test cases themselves have not been changed.


vmihailenco:
Thanks :+1:

PR #975, opened 2024-04-07
feat: apply ModelTableExpr for SELECT query

### LINKED ISSUE #968: inconsistent behavior for ModelTableExpr (select <--> create/insert/update/delete)
bun@v1.1.17

bun allows using ModelTableExpr to rename table, insert/update could operate on renamed table, but query_select doesn't support to select from it.

I raised https://github.com/uptrace/bun/issues/965 but it was closed.

From the tracing log, select ignored the parameter of ModelTableExpr when constructing the sql statement.
```terminal
[bun]  08:16:07.856   CREATE TABLE          1.421ms  CREATE TABLE renamed_table_name ("id" INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, "name" VARCHAR)
[bun]  08:16:07.856   INSERT                  168µs  INSERT INTO renamed_table_name ("name") VALUES ('admin') RETURNING "id"
[bun]  08:16:07.856   UPDATE                   86µs  UPDATE renamed_table_name SET "name" = 'root' WHERE (id = 1)
[bun]  08:16:07.857   SELECT                   32µs  SELECT "user"."id", "user"."name" FROM renamed_table_name WHERE (id = 1)     *sqlite.Error: SQL logic error: no such column: user.id (1) 
Fail to select user  SQL logic error: no such column: user.id (1)
[bun]  08:16:07.857   DELETE                   80µs  DELETE FROM renamed_table_name WHERE (id = 1 )
```

```go
func DemoBunOperation() {
	sqldb, err := sql.Open(sqliteshim.ShimName, "file::memory:?cache=shared")
	if err != nil {
		panic(err)
	}

	db.AddQueryHook(bundebug.NewQueryHook(
		bundebug.WithVerbose(true),
		bundebug.FromEnv("BUNDEBUG"),
	))

	ctx := context.Background()
	type User struct {
		bun.BaseModel

		ID   int64 `bun:",pk,autoincrement"`
		Name string
	}

	renamedTable := "renamed_table_name"
	db.NewCreateTable().Model((*User)(nil)).ModelTableExpr(renamedTable).Exec(ctx)
	user := &User{Name: "admin"}
	db.NewInsert().Model(user).ModelTableExpr(renamedTable).Exec(ctx)
	user.Name = "root"
	_, err = db.NewUpdate().Model(user).ModelTableExpr(renamedTable).Where("id = ?", user.ID).Exec(ctx)
	if err != nil {
		fmt.Println("Fail to update user ", err.Error())
	}
	readUser := new(User)
	err = db.NewSelect().Model(readUser).ModelTableExpr(renamedTable).Where("id = ?", user.ID).Scan(ctx)
	if err != nil {
		fmt.Println("Fail to select user ", err.Error())
	}
	_, err = db.NewDelete().Model(user).ModelTableExpr(renamedTable).Where("id = ? ", user.ID).Exec(ctx)
	if err != nil {
		fmt.Println("Fail to delete user ", err.Error())
	}
}
```

JunNishimura:
## Overview 
This PR fixes #968. 

The ModelTableExpr method in the existing implementation simply overwrites the table name in the FROM clause, but not the table name used as the column name prefix in the SELECT clause (such as SELECT users.name). Therefore, the Insert, Update, and Delete processes were fine, but the Select process was not working as expected.

### 1. add Table Parser
Extract table names and aliases from queries passed as arguments to ModelTableExpr method.

### 2. Update columns builder of select query
Consider the table name specified in ModelTableExpr when constructing columns in the SELECT clause.

### DIFF
--- a/internal/dbtest/query_test.go
+++ b/internal/dbtest/query_test.go
@@ -1045,6 +1045,12 @@ func TestQuery(t *testing.T) {
 			}
 			return db.NewInsert().Model(new(Model))
 		},
+		func(db *bun.DB) schema.QueryAppender {
+			return db.NewSelect().Model(new(Model)).ModelTableExpr("renamed")
+		},
+		func(db *bun.DB) schema.QueryAppender {
+			return db.NewSelect().Model(new(Model)).ModelTableExpr("renamed AS r")
+		},
 	}
 
 	timeRE := regexp.MustCompile(`'2\d{3}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}(\.\d+)?(\+\d{2}:\d{2})?'`)
--- a/internal/dbtest/query_test.go
+++ b/internal/dbtest/query_test.go
@@ -1545,11 +1545,17 @@ func TestQuery(t *testing.T) {
 				return db.NewInsert().Model(new(Model))
 			},
 		},
-		func(db *bun.DB) schema.QueryAppender {
-			return db.NewSelect().Model(new(Model)).ModelTableExpr("renamed")
+		{
+			id: 167,
+			query: func(db *bun.DB) schema.QueryAppender {
+				return db.NewSelect().Model(new(Model)).ModelTableExpr("renamed")
+			},
 		},
-		func(db *bun.DB) schema.QueryAppender {
-			return db.NewSelect().Model(new(Model)).ModelTableExpr("renamed AS r")
+		{
+			id: 168,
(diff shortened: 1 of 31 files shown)


vmihailenco:
Some time ago we/I decided to not tackle this problem so we don't have to implement table parser.

While I appreciate the work, I still think that this library should not attempt to parse/manipulate SQL queries.

PR #978, opened 2024-04-09
test: add unit test for parser

JunNishimura:
I added unit tests for parser package because it is an important feature that affects the entire Bun.


vmihailenco:
Thanks!

PR #995, opened 2024-06-02
Support Oracle dialect

mattn:
Hi, I'm author of [go-oci8](https://github.com/mattn/go-oci8). Now I added support for Oracle dialect. Currently, this todo app is working on my environment with small changes for driver name.

https://github.com/mattn/todo-webapp


vmihailenco:
@mattn nice, thanks! Any chance we can start running tests for Oracle [here](https://github.com/uptrace/bun/blob/master/internal/dbtest/db_test.go#L46-L54)?

Also, if you can provide any examples, I will add the to the docs next to the other [dialects](https://bun.uptrace.dev/guide/drivers.html).

mattn:
Thanks. This is an exmaple app using uptrace/bun and go-oci8.

https://github.com/mattn/todo-webapp/blob/oci8/main.go

What example do you prefer?

vmihailenco:
@mattn added an example to the [doc](https://bun.uptrace.dev/guide/drivers.html#oracle).

Issue #992, opened 2024-05-28
m2m relations with composite keys


ygabuev:
I am trying to have a many-to-many relation on models with composite primary keys. Here is a concrete example:

```sql
CREATE TABLE orders (id int, shop_id int);
CREATE TABLE items (id int, shop_id int);
CREATE TABLE orders_to_items (shop_id int, order_id int, item_id int);

INSERT INTO orders VALUES (1, 22);
INSERT INTO items VALUES (11, 22);
INSERT INTO orders_to_items VALUES (22, 1, 11);
```

```go
package main

import (
	"context"
	"database/sql"
	"os"

	"github.com/uptrace/bun"
	"github.com/uptrace/bun/dialect/pgdialect"
	"github.com/uptrace/bun/driver/pgdriver"
	"github.com/uptrace/bun/extra/bundebug"
)

type Order struct {
	ID     int    `bun:",pk"`
	ShopID int    `bun:",pk"`
	Items  []Item `bun:"m2m:orders_to_items,join:Order=Item"`
}

type Item struct {
	ID     int `bun:",pk"`
	ShopID int `bun:",pk"`
}

type OrderToItem struct {
	bun.BaseModel `bun:"table:orders_to_items"`

	ShopID int `bun:""`

	OrderID int    `bun:""`
	Order   *Order `bun:"rel:belongs-to,join:order_id=id,join:shop_id=shop_id"`
	ItemID  int    `bun:""`
	Item    *Item  `bun:"rel:belongs-to,join:item_id=id,join:shop_id=shop_id"`
}

func main() {
	sqldb := sql.OpenDB(pgdriver.NewConnector(
		pgdriver.WithDSN(os.Getenv("DB_URL")),
		pgdriver.WithPassword(os.Getenv("DB_PASSWORD")),
	))

	db := bun.NewDB(sqldb, pgdialect.New())
	db.AddQueryHook(bundebug.NewQueryHook(bundebug.FromEnv("BUNDEBUG")))
	db.RegisterModel((*OrderToItem)(nil))

	var orders []Order

	err := db.NewSelect().
		Model(&orders).
		Relation("Items").
		Scan(context.Background())
	if err != nil {
		panic(err)
	}
}
```

[...]

PR #996, opened 2024-06-05
Fix M2M relations for models with composite keys

ygabuev:
This PR adds support for M2M relations with composite keys. This feature was probably supposed to be working previously, but was not tested and [turned out](https://github.com/uptrace/bun/issues/992) to be buggy. Existing functionality is not affected, according to the tests.

Commits:
- **fix(m2m join): don't query join columns**
- **fix(m2m): enforce ordering of PKs of base table**
- **fix(m2m): remove unneeded appending to struct key**
- **fix(m2m): correctly infer which columns belong to which model**
- **feat: add test for M2M relations with composite keys**

Closes https://github.com/uptrace/bun/issues/992.

### DIFF
--- a/internal/dbtest/orm_test.go
+++ b/internal/dbtest/orm_test.go
@@ -32,6 +32,7 @@ func TestORM(t *testing.T) {
 		{testM2MRelationExcludeColumn},
 		{testRelationBelongsToSelf},
 		{testCompositeHasMany},
+		{testCompositeM2M},
 	}
 
 	testEachDB(t, func(t *testing.T, dbName string, db *bun.DB) {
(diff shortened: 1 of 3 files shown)


vmihailenco:
>dialect does not support the (a, b) in ((v1, v2), ...) syntax for lists of tuple-like objects.

I wonder have you discarded the `VALUES (x, y), (foo, bar)` trick from the stackoverflow to make MSSQL work?

vmihailenco:
@ygabuev thanks. I've skipped MSSQL tests for now.

>e.g. along [these lines](https://stackoverflow.com/a/8006966/7500634), but this would result in lengthier queries and thus negatively affects all the other SQL backends.

I think this is totally fine, but let's not block this PR because of MSSQL.

PR #983, opened 2024-04-27
fix: has-many join failing with pointer key fields (#950)

jeffreydwalter:

PR #1000, opened 2024-06-18
Fix matching in has-many relation

ygabuev:


vmihailenco:
Might be fixed by this https://github.com/uptrace/bun/pull/983

Issue #998, opened 2024-06-14
Is it possible to insert a comment before or after a sql query?


pragkent:
Is it possible to insert a comment before or after a sql query? We have a database middleware which parse hint from comment and route the request to the real backend mysql instances.

The query is something like:
```sql
/* shard = shard-00 */ select * from users where id = 100;
```

PR #1002, opened 2024-07-01
feat: add mechanism for prefixing queries with a comment string

kylemcc:
This change proposes a new `Comment(string)` method and `ContextWithComment(context.Context, string)` function for prepending a comment to a SQL query.

Here's simple example:

```go
// using the query API
err := db.NewSelect().
    Model(&products).
    Where("description like ?", d).
    Comment("customer:"+customerID).
    Scan(ctx)

// using a context
qctx := bun.ContextWithComment(ctx, "customer:"+customerID)
err := db.NewSelect().
    Model(&products).
    Where("description LIKE ?", d).
    Scan(qctx)
```
This would produce a query that looks like:

```sql
/*customer:01J1P97P3AP7V287B57P7055CZ*/ SELECT * FROM products WHERE description LIKE '%red%'
```

This may warrant a bit of discussion first, but I wanted to include the code with the proposal. I have had this code running in production since April with no problems. It appears that others may be interested in this as well: https://github.com/uptrace/bun/issues/998

My immediate use case for this is to add a comment to a query with tracing/debug information that will show up in the process list/`pg_stat_activity`, but there are certainly other uses for this functionality.

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -74,6 +74,7 @@ var (
 // QueryBuilder is used for common query methods
 type QueryBuilder interface {
 	Query
+	Comment(c string) QueryBuilder
 	Where(query string, args ...interface{}) QueryBuilder
 	WhereGroup(sep string, fn func(QueryBuilder) QueryBuilder) QueryBuilder
 	WhereOr(query string, args ...interface{}) QueryBuilder
@@ -103,6 +104,8 @@ type baseQuery struct {
 	modelTableName schema.QueryWithArgs
 	tables         []schema.QueryWithArgs
 	columns        []schema.QueryWithArgs
+
+	comment string
 
 	flags internal.Flag
 }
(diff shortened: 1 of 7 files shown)


query_base.go
```diff
@@ -146,6 +149,32 @@ func (q *baseQuery) GetTableName() string {
 	return ""
 }
 
+func (q *baseQuery) setComment(c string) {
+	q.comment = c
+}
+
+func (q *baseQuery) appendComment(b []byte) []byte {
```
vmihailenco, line 156: Could you please rework this to accept `ctx context.Context` and then extract the comment from the context in this function?

Then we can remove `commentFromContext`.

vmihailenco:
@kylemcc This looks useful, but I have a minor refactoring request. Please take a look when you have time.

query_base.go
kylemcc: Thanks for the review @vmihailenco! Updated as you requested and rebased on top of master.

query_base.go
```diff
@@ -146,6 +149,38 @@ func (q *baseQuery) GetTableName() string {
...
+}
+
+func (q *baseQuery) appendComment(ctx context.Context, b []byte) []byte {
+	// check if a comment was explicitly set on the query using the .Comment("...") api
+	if q.comment == "" {
+		// if not, check if one exists on the incoming context
+		if c, _ := ctx.Value(queryCommentCtxKey{}).(string); c == "" {
+			return b
+		} else {
+			q.comment = c
```
vmihailenco, line 163: Can you please change this so it does not modify the `baseQuery`? So the following is possible:

```
q := NewSelectQuery()
q.AppendQuery(ctxWithComment1) // one comment
q.AppendQuery(ctxWithComment2) // another comment
```

Also, one might think that both comments must be appended: one from the query and one from the context. I will leave it to you to decide what to do in such case.

vmihailenco: If multiple comments are allowed, then we can also consider supporting this: `NewSelect().Comment("comment1").Comment("comment2")`.

kylemcc: > Can you please change this so it does not modify the `baseQuery`? So the following is possible:
> 
> ```
> q := NewSelectQuery()
> q.AppendQuery(ctxWithComment1) // one comment
> q.AppendQuery(ctxWithComment2) // another comment
> ```
> 
> Also, one might think that both comments must be appended: one from the query and one from the context. I will leave it to you to decide what to do in such case.

Ah, yes, good catch. Didn't think about the `AppendQuery` use case.

> If multiple comments are allowed, then we can also consider supporting this: NewSelect().Comment("comment1").Comment("comment2").

This seems reasonable, too. Let me experiment with the formatting a bit to see what looks best.

vmihailenco:
@kylemcc any chance you can finish this?

kylemcc:
> @kylemcc any chance you can finish this?

Apologies @vmihailenco, got busy with other things. Let me see if I can get this cleaned up today or tomorrow

PR #1006, opened 2024-07-12
Perf/cha 11  perf memory

marvinnubit:


vmihailenco:
We pre-allocate some memory for the query to *improve* performance. We'd need some benchmark to change this.

PR #1009, opened 2024-07-16
fix: allow Limit() without Order() with MSSQL

### LINKED ISSUE #811: Erro dialect mssql witch NEXT in the FETCH

CL-Jeremy:
Implements workaround from https://stackoverflow.com/a/36156953 for all backends supporting `feature.OffsetFetch` from ISO/ANSI SQL:2008, which specifies `ORDER BY` as a requirement. This should be tested further if more backends are to be added (such as Oracle in #995 – this feature is not enable there, and there seems to be an impact noted by https://stackoverflow.com/q/48620590 and https://stackoverflow.com/q/59864964).

Closes #811

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -541,7 +541,6 @@ func (q *SelectQuery) appendQuery(
 		// MSSQL: allows Limit() without Order() as per https://stackoverflow.com/a/36156953
 		if q.limit > 0 && len(q.order) == 0 && fmter.Dialect().Name() == dialect.MSSQL {
 			b = append(b, "0 AS _temp_sort, "...)
-			q.order = []schema.QueryWithArgs{schema.UnsafeIdent("_temp_sort")}
 		}
 
 		b, err = q.appendColumns(fmter, b)
--- a/query_select.go
+++ b/query_select.go
@@ -538,11 +538,12 @@ func (q *SelectQuery) appendQuery(
 	if count && !cteCount {
 		b = append(b, "count(*)"...)
 	} else {
-		// Allows Limit() without Order() with MSSQL as per https://stackoverflow.com/a/36156953
-		if q.limit > 0 && fmter.Dialect().Features().Has(feature.OffsetFetch) && len(q.order) == 0 {
+		// MSSQL: allows Limit() without Order() as per https://stackoverflow.com/a/36156953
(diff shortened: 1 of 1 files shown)


query_select.go
```diff
@@ -538,6 +538,11 @@ func (q *SelectQuery) appendQuery(
 	if count && !cteCount {
 		b = append(b, "count(*)"...)
 	} else {
+		// Allows Limit() without Order() with MSSQL as per https://stackoverflow.com/a/36156953
+		if q.limit > 0 && fmter.Dialect().Features().Has(feature.OffsetFetch) && len(q.order) == 0 {
+			b = append(b, "0 AS _temp_sort, "...)
+			q.order = []schema.QueryWithArgs{schema.UnsafeIdent("_temp_sort")}
```
vmihailenco, line 544: This is query generation and it should not change the original `SelectQuery`, because someone may be reusing them.

You probably need to add another block of code to generate order without changing the `SelectQuery`.

CL-Jeremy: Do you mean that if I have a `SelectQuery` in a variable `a` and then do `b := a.Limit(5)`, the value of `a` would change, and that would be a problem? Isn't that also the case without the proposed change? I cannot seem to find an analog for this case to compare to. Do you mean something like `ORDER BY` after `LIMIT`, or is there another case where there could be something appended to a select query after `Limit()`? An example would be much appreciated.

vmihailenco: @CL-Jeremy I've fixed this myself. See the commits in this branch.

PR #1010, opened 2024-07-22
fix: gracefully handle empty hstore in pgdialect

0rax:
In the current implementation of `hstore` in `github.com/uptrace/bun/dialect/pgdialect`, an empty `hstore` is encoded properly in the database as `''` but can not be retrieved.

It is handled properly for a `var md map[string]string = nil` when `NULL` is inserted in database and an empty map is returned when querying back the record, but when using a `map[string]string{}` it will insert `''` (which is expected) but will fail with the following error:

```
sql: Scan error on column index 8, name "metadata": bun: can't parse hstore: ""
```

This PR make sure an empty `hstore` is handled gracefully in the `hstoreParser` as well as adds test cases for this issues in `pgdialect` itself (for the parser AND the appender) as well as in the `internal/dbtest` package.

This fix is relatively crucial to us, as we tend to store user defined metadata in `hstore` and it may happen that a user removes all key value pairs from there. Currently, we have to check if the changes made to a metadata field makes it empty and force the field to be null, being able to handle empty hstore gracefully at the database driver level would make our life quite easier here as it seems like a legitimate use of that type.

### DIFF
--- a/dialect/pgdialect/append_test.go
+++ b/dialect/pgdialect/append_test.go
@@ -1,0 +1,39 @@
+package pgdialect
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/uptrace/bun/schema"
+)
+
+func TestHStoreAppender(t *testing.T) {
+	tests := []struct {
+		input      map[string]string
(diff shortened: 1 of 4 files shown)


vmihailenco:
Thanks for fixing this.

PR #1019, opened 2024-09-03
fix: sql.NullString mistaken as custom struct

CL-Jeremy:
Tested with this concise official example with connection setup changed to PostgreSQL:

https://github.com/uptrace/bun/blob/6c83b05f27c70bf3aff7df90886743b7e313927e/example/fixture/main.go#L34-L55

Without this fix, `sql.NullString` gets treated as a custom struct and then mapped to JSONB, resulting in failed insertion operations after the table has been created.

(Also removed currently unused, extraneous types.)

### DIFF
--- a/dialect/pgdialect/sqltype.go
+++ b/dialect/pgdialect/sqltype.go
@@ -1,6 +1,7 @@
 package pgdialect
 
 import (
+	"database/sql"
 	"encoding/json"
 	"net"
 	"reflect"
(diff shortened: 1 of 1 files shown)


CL-Jeremy:
Of course, there could be more tests added to avoid this type of mistakes systematically. Just a bit unsure where to put such tests (for PGSQL specifically, or should there be a general, complicated test, e. g. loading fixtures, enabled for all backends?)

PR #1028, opened 2024-10-14
Add support type for net/netip.Addr and net/netip.Prefix

Aoang:


vmihailenco:
@Aoang the build fails

PR #1048, opened 2024-11-02
MySQL/MariaDB: add support for ORDER BY and LIMIT clauses in UPDATE/DELETE statements

capcom6:
This PR adds support for `ORDER BY` and `LIMIT` clauses in `DELETE` statements specifically for MySQL and MariaDB databases.

**Changes:**
- Extract `orderQuery` and `limitOffsetQuery` from `SelectQuery`, see `query_base.go` and `query_select.go`
- Include `orderQuery` and `limitOffsetQuery` in `DeleteQuery`, see `query_delete.go`
- Implement `Order` and `Limit` methods on `DeleteQuery`, see `query_delete.go`
- Add dbtest cases 168-170, see `query_test.go`

**Example usage:**
```go
db.NewDelete().Model(new(Model)).WherePK().Order("id")
db.NewDelete().Model(new(Model)).WherePK().Order("id").Limit(1)
```

This feature enhances the flexibility of `DELETE` operations in MySQL and MariaDB, allowing for more precise control over which rows are deleted and in which order.

**Questions:**
I have some questions about which I'm unsure:

1. PR #1009 has allowed `Limit` without `Order` for MSSQL. As a side effect, it has introduced a dependency between these parts of the query, which in my PR is separated. Before #1009, my implementations fully moved `ORDER` and `LIMIT...OFFSET` parts into reusable and independent parts. Now I had to "override" `appendQuery` in `SelectQuery` to support this MSSQL behavior.

2. Because only MySQL/MariaDB supports syntax with `LIMIT` and `ORDER BY` for `DELETE`, I'm unsure where to return an error for other dialects. For now, I set `err` in `Order`, `OrderExpr`, and `Limit` methods of `DeleteQuery`.

3. This syntax can't be used with multiple tables. The only place to validate this that I found is `AppendQuery`. Is this correct?

**Additional context:**

- For now, I implemented this feature by directly checking the dialect and not by features because, as far as I know, only MySQL/MariaDB supports this syntax..

**References:**
- [MySQL 5.7](https://dev.mysql.com/doc/refman/5.7/en/delete.html#idm45572293043200)
- [MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/delete.html#idm45599522668000)
- [MySQL 9.1](https://dev.mysql.com/doc/refman/9.1/en/delete.html#idm46701568070240)
- [MariaDB](https://mariadb.com/kb/en/delete/)

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -6,6 +6,8 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/uptrace/bun/dialect"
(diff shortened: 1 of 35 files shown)


j2gg0s:
Thanks for your work.

I only have a rough understanding of the background for now and don’t have any well-developed ideas yet.
But I have some simple ideas:
1. MySQL's Update also support ORDER BY. Maybe we should add a [feature flag](https://github.com/uptrace/bun/blob/master/dialect/feature/feature.go#L8) to represent it.
2. Where ORDER is supported, LIMIT&OFFSET is always supported as well, so we might not need to split it into two class in our code. 
3. Let

capcom6:
Thanks for your comment.

I apologize for not mentioning the background earlier. The syntax `DELETE ... ORDER BY ... LIMIT ... RETURNING ...` can be used to implement the [Transactional outbox](https://microservices.io/patterns/data/transactional-outbox.html) pattern, allowing for reading messages in a single query with deletion and specified order. Additionally, `ORDER BY...` in `DELETE` and `UPDATE` statements can be helpful when working with relations to control the order of changes and prevent foreign key constraint violations.

Regarding your ideas:
1. I've added support for this syntax to the `UPDATE` statement and introduced feature flags for syntax support in both `UPDATE` and `DELETE`.
2. Agreed. I've merged the structs and moved #1009 from `SelectQuery` to `orderLimitOffsetQuery`.

P.S. For now, I've created distinct commits for changes to simplify the review process. I'll edit the commit history later.

query_select.go
```diff
@@ -783,28 +729,7 @@ func (q *SelectQuery) appendTables(fmter schema.Formatter, b []byte) (_ []byte,
 }
 
 func (q *SelectQuery) appendOrder(fmter schema.Formatter, b []byte) (_ []byte, err error) {
```
j2gg0s, line 731: We don't need wrap this method.

query_base.go
```diff
@@ -1352,3 +1354,113 @@ func (ih *idxHintsQuery) bufIndexHint(
...
+//------------------------------------------------------------------------------
+
+type orderLimitOffsetQuery struct {
+	order []schema.QueryWithArgs
+
+	limit  int32
+	offset int32
+}
+
+func (olfq *orderLimitOffsetQuery) addOrder(orders ...string) {
```
j2gg0s, line 1367: Why use olfq instead of q?

query_base.go
```diff
@@ -1352,3 +1354,113 @@ func (ih *idxHintsQuery) bufIndexHint(
...
+
+	// MSSQL: allows Limit() without Order() as per https://stackoverflow.com/a/36156953
+	if olfq.limit > 0 && fmter.Dialect().Name() == dialect.MSSQL {
+		return append(b, " ORDER BY _temp_sort"...), nil
+	}
+
+	return b, nil
+}
+
+func (olfq *orderLimitOffsetQuery) addLimit(n int) {
```
j2gg0s, line 1425: set is better than add

query_base.go
```diff
@@ -1352,3 +1354,113 @@ func (ih *idxHintsQuery) bufIndexHint(
...
+	olfq.limit = int32(n)
+}
+
+func (olfq *orderLimitOffsetQuery) addOffset(n int) {
```
j2gg0s, line 1429: set is better than add

query_delete.go
```diff
@@ -120,11 +122,36 @@ func (q *DeleteQuery) WhereAllWithDeleted() *DeleteQuery {
 	return q
 }
 
+func (q *DeleteQuery) Order(orders ...string) *DeleteQuery {
+	if !q.hasFeature(feature.DeleteOrderLimit) {
+		q.err = errors.New("bun: order is not supported for current dialect")
```
j2gg0s, line 127: if err, just return immediately.

query_base.go
capcom6: Good question! Probably inspired by `func (ih *idxHintsQuery)`. Fixed.

query_base.go
capcom6: Fixed.

query_base.go
capcom6: Fixed.

query_delete.go
capcom6: Fixed.

query_select.go
capcom6: Oh yeah, it was used for MSSQL before the structures were merged.

j2gg0s:
Looking forward to you squashing the commits and officially submitting the changes.

PR #1049, opened 2024-11-07
feat: Enhance Apply method to accept multiple functions

Tiscs:
This pull request includes updates to the `Apply` method across multiple query types to support variadic function arguments, allowing multiple functions to be applied in sequence, these changes **will not break** existing calls.

```go
type UsersRepo interface {
    // Implemention:
    //
    // var users []*models.User
    // query := r.bdb.NewSelect().Model(&users).Apply(fns...)
    // return query.ScanAndCount(ctx)
    //
    // Example 1:
    //
    //   // find users 
    //   users, _, err := usersRepo.FindUsers(ctx,
    //       func(sq *bun.SelectQuery) *bun.SelectQuery {
    //           return sq.Where("age > ?", 59).Order("age DESC")
    //       },
    //       func(sq *bun.SelectQuery) *bun.SelectQuery {
    //           return sq.Offset(0).Limit(100)
    //       },
    //   )
    FindUsers(ctx context.Context, fns ...func(*bun.SelectQuery) *bun.SelectQuery) ([]*models.User, int, error)
}
```

BTW, some comments were fixed.

### DIFF
--- a/query_column_add.go
+++ b/query_column_add.go
@@ -42,9 +42,12 @@ func (q *AddColumnQuery) Err(err error) *AddColumnQuery {
 	return q
 }
 
-func (q *AddColumnQuery) Apply(fn func(*AddColumnQuery) *AddColumnQuery) *AddColumnQuery {
-	if fn != nil {
-		return fn(q)
+// Apply calls each function in fns, passing the AddColumnQuery as an argument.
+func (q *AddColumnQuery) Apply(fns ...func(*AddColumnQuery) *AddColumnQuery) *AddColumnQuery {
+	for _, fn := range fns {
+		if fn != nil {
+			q = fn(q)
+		}
 	}
 	return q
 }
--- a/query_column_drop.go
+++ b/query_column_drop.go
@@ -40,9 +40,12 @@ func (q *DropColumnQuery) Err(err error) *DropColumnQuery {
 	return q
 }
 
-func (q *DropColumnQuery) Apply(fn func(*DropColumnQuery) *DropColumnQuery) *DropColumnQuery {
-	if fn != nil {
-		return fn(q)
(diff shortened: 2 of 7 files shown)


j2gg0s:
Thanks for your contribution!

However, I think Query is chainable, so could we achieve the same effect with the following approach?
```go
db.NewSelect().Model(new(User)).Apply(fn1).Apply(fn2)
```

Tiscs:
> Thanks for your contribution!
> 
> However, I think Query is chainable, so could we achieve the same effect with the following approach?
> 
> ```go
> db.NewSelect().Model(new(User)).Apply(fn1).Apply(fn2)
> ```

Thanks for your response.

That's right, the query can be implemented in your way.

But in my case, I build the query by passing in variadic parameters, so I need to do it like this:

```go
q := db.NewSelect().Model(new(User))
for _, fn := range fns {
    fn(q) // or `q = fn(q)`, queries are modified in-place
}
q.Scan(ctx)
```

So, I think chainable queries are great, and applying multiple at once is also useful.

query_column_add.go
```diff
@@ -42,9 +42,10 @@ func (q *AddColumnQuery) Err(err error) *AddColumnQuery {
 	return q
 }
 
-func (q *AddColumnQuery) Apply(fn func(*AddColumnQuery) *AddColumnQuery) *AddColumnQuery {
-	if fn != nil {
-		return fn(q)
+// Apply calls each function in fns, passing the AddColumnQuery as an argument.
+func (q *AddColumnQuery) Apply(fns ...func(*AddColumnQuery) *AddColumnQuery) *AddColumnQuery {
+	for _, fn := range fns {
+		q = fn(q)
```
j2gg0s, line 48: I’m concerned that some existing code might be passing in nil. Could we add compatibility for that?

PR #1050, opened 2024-11-07
github/action: add action stale

j2gg0s:


vmihailenco:
@j2gg0s can we have some tag to mark some issues as not stale? It would be brutal to require users to bump the thread each week so it is not closed...

j2gg0s:
> @j2gg0s can we have some tag to mark some issues as not stale? It would be brutal to require users to bump the thread each week so it is not closed...

@vmihailenco 
- we can use `no stalebot` or `help wanted`, [stale.yml#L37](https://github.com/uptrace/bun/blob/master/.github/workflows/stale.yml#L37)
- A issue/pr labeled as stale will have the stale label removed and be granted another 30-day grace period once it becomes active again.

PR #1056, opened 2024-11-11
Fix #981: unique tag in embeded structure

### LINKED ISSUE #981: In v1.2, about 'unique' tag option in embedded structures
After updated to v1.2.1 today, I noticed that my unit tests which used to pass are now failing. It seems like the unique tag option in nested structures has become ineffective, and this issue is present in v1.2.0 as well.

My Go version is v1.22.0, with PostgreSQL 15, and tested in macOS 14.

More detail in https://github.com/cupogo/andvari/blob/main/stores/pgx/z0_main_test.go

```go
type Clause struct {
	comm.BaseModel `bun:"table:cms_clause,alias:c" json:"-"`

	comm.DefaultModel

	ClauseBasic
	comm.MetaField
}

type ClauseBasic struct {
	Slug  string   `bun:"slug,notnull,type:name,unique" json:"slug" `
	Text  string   `bun:"text,notnull,type:text" form:"text" json:"text" `
	Cates []string `bun:"cates,notnull,type:jsonb" form:"cats" json:"cats" `
}

```

WX20240423-192758@2x

I've searched through the most recent change history and still couldn't pinpoint the specific reason.

liut:
‌‌In the case where there were many changes in version 1.17 + 1 commit, subsequent versions did not work well with a large number of embeds in my scenario. For this reason, I made minor revisions to make it work. When a unique is detected in the subtable, it is set as the only unique. Generally, Unique tags are not used too much. In my scenario, it just solved the problem.

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -174,6 +174,10 @@ func (t *Table) processFields(typ reflect.Type, canAddr bool) {
 				}
 			}
 
+			if subtable.Unique != nil && t.Unique == nil {
+				t.Unique = subtable.Unique
+			}
+
 			continue
 		}
 


schema/table.go
```diff
@@ -174,6 +174,10 @@ func (t *Table) processFields(typ reflect.Type, canAddr bool) {
 				}
 			}
 
+			if subtable.Unique != nil && t.Unique == nil {
+				t.Unique = subtable.Unique
+			}
+
```
j2gg0s, line 180: I think we should add the Unique constraint when adding the embeded field later Line 246

j2gg0s (commented):
Good job

j2gg0s:
We hadn’t previously defined the behavior for when embed and Unique appear together, so I’ve created some test cases.

```go
        t.Run("embedWithUnique", func(t *testing.T) {
                type Perms struct {
                        View          bool
                        Create        bool
                        UniqueID      int `bun:",unique"`
                        UniqueGroupID int `bun:",unique:groupa"`
                }

                type Role struct {
                        Foo Perms `bun:"embed:foo_"`
                        Perms
                }

                table := tables.Get(reflect.TypeOf((*Role)(nil)))
                require.Nil(t, table.StructMap["foo"])
                require.Nil(t, table.StructMap["bar"])

                fooView, ok := table.FieldMap["foo_view"]
                require.True(t, ok)
                require.Equal(t, []int{0, 0}, fooView.Index)

                barView, ok := table.FieldMap["view"]
                require.True(t, ok)
                require.Equal(t, []int{1, 0}, barView.Index)

                require.Equal(t, 3, len(table.Unique))
                require.Equal(t, 2, len(table.Unique[""]))
                require.Equal(t, "foo_unique_id", table.Unique[""][0].Name)
                require.Equal(t, "unique_id", table.Unique[""][1].Name)
                require.Equal(t, 1, len(table.Unique["groupa"]))
                require.Equal(t, "unique_group_id", table.Unique["groupa"][0].Name)
                require.Equal(t, 1, len(table.Unique["foo_groupa"]))
                require.Equal(t, "foo_unique_group_id", table.Unique["foo_groupa"][0].Name)
        })
```

PR #1059, opened 2024-11-13
MariaDB: add support for RETURNING clause in DELETE statements

capcom6:
This PR adds support for `RETURNING` clauses in `DELETE` statements specifically for MariaDB databases. MariaDB has supported the `RETURNING` clause in `DELETE` statements since version 10.0.5.

**Changes:**
- Introduce `feature.DeleteReturning` because the existing `feature.Returning` is used in `UpdateQuery`, but MySQL/MariaDB don't support `RETURNING` clause in `UPDATE` statements
- Add `feature.DeleteReturning` to all dialects already having `feature.Returning` for backward compatibility
- Add `feature.DeleteReturning` to `mysql` dialect for MariaDB 10.0.5 and later
- Replace `feature.Returning` with `feature.DeleteReturning` in `DeleteQuery`
- Add check for `feature.DeleteReturning` in `DeleteQuery.Returning` method because previously, `Returning` could be used for the `mysql` dialect but didn't receive any error or result

**Example usage:**
```go
db.NewDelete().Model(new(Model)).WherePK().Returning("*")
```

**Notes:**
- MariaDB doesn't support `RETURNING` clause for multi-table deletes but I haven't add any check for that because I don't know is any other database supports this.

**Discussion:**
- Not directly related to this PR: There is `feature.InsertReturning`, but `InsertQuery.tryLastInsertID` checks `feature.Returning`. Is this correct?
- After this PR, `feature.Returning` will be used only in `UpdateQuery`. I propose renaming it to `feature.UpdateReturning` to clarify its usage. This renaming has not been implemented yet.

**References:**
- [MDEV-3814](https://jira.mariadb.org/browse/MDEV-3814)
- [MariaDB 10.0.5 Release Notes](https://mariadb.com/kb/en/mariadb-1005-release-notes/)


dialect/oracledialect/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/uptrace/bun/dialect/oracledialect
 
-go 1.22.0
+go 1.23
```
j2gg0s, line 3: Dont upgrade go version, we support go >= 1.22

query_delete.go
```diff
@@ -163,6 +163,11 @@ func (q *DeleteQuery) Limit(n int) *DeleteQuery {
 //
 // To suppress the auto-generated RETURNING clause, use `Returning("NULL")`.
 func (q *DeleteQuery) Returning(query string, args ...interface{}) *DeleteQuery {
+	if !q.hasFeature(feature.DeleteReturning) {
+		q.err = errors.New("bun: returning is not supported for current dialect")
+		return q
+	}
+
```
j2gg0s, line 170: For dialects that do not support DELETE RETURNING.
Before: We remained slient on misuse and ignore it when append query.
After: We return error on misuse.

I dont think current solution is correct, but i dont think we should break it,
as that might casue potential failures in upper-level application.

dialect/oracledialect/go.mod
capcom6: But it was already broken in https://github.com/uptrace/bun/pull/926 and merged into `master`: all `go.mod` files were updated and `1.22` was removed from `.github/workflows/build.yml`

j2gg0s: Thanks, i just saw it.

query_delete.go
capcom6: On the other hand, I was very discouraged when I tried to use `Returning()` with `DeleteQuery` and received an empty result. I propose adding a note to the method's documentation, something like: "If the dialect does not support DELETE RETURNING, an empty result will be returned without an error."

j2gg0s: I dont have a strong preference, let @vmihailenco decide.

vmihailenco: I think returning an error might be annoying in some cases, but way easier to debug...

PR #1067, opened 2024-11-20
fix: update oracledialect/version.go in release.sh

mattn:


j2gg0s:
Good Job

Do you mind if we directly change it to the corresponding directory under dialect for convenience?

mattn:
Yes, please.

PR #1069, opened 2024-11-21
ci: use find replace hardcode when modify func Version

j2gg0s:


scripts/release.sh
```diff
@@ -53,26 +53,23 @@ if [[ "$(uname)" == "Darwin" ]]; then
...
-    sed -i "" "s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./dialect/mysqldialect/version.go
-    sed -i "" "s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./dialect/mssqldialect/version.go
-    sed -i "" "s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./dialect/pgdialect/version.go
-    sed -i "" "s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./dialect/sqlitedialect/version.go
-    sed -i "" "s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./dialect/oracledialect/version.go
+    for file in $(find . -type f -name 'version.go')
+    do
+        sed -i "" "/func Version() string/{n;s/\(return \)\"[^\"]*\"/\1\"${TAG#v}\"/;}" ${file}
+    done
     sed -i "" "s/\(\"version\": \)\"[^\"]*\"/\1\"${TAG#v}\"/" ./package.json
```
vmihailenco, line 60: I believe we can use this version `sed -i` everywhere, i.e. it seems to be supported on Ubuntu/Linux.

j2gg0s: @vmihailenco 

`sed -i ""` is feature of BSD, we must use `sed -i` in GNU

vmihailenco: Alright, let's keep it as is then :+1:

vmihailenco: Alright, let's keep it as is then :+1:

Issue #1045, opened 2024-10-31
Default values being ignored on NewInsert?


fhendrikx:
Hi,

I'm new to Bun, hopefully you can put me right.

When inserting a record using NewInsert(), it seems to ignore my request to use the default value.

For example, the model contains a "State" integer:

     State  int  `bun:",nullzero,notnull,default:4"`

But when creating the record and then inserting it:

```
        users := []*User{
                {Username: "test 1", Name: "testing"},
                {Username: "test 2", Name: "testing"},
        }

        db.NewInsert().Model(&users).Exec(ctx)
```

The database complains:

 `  pgdriver.Error: ERROR: null value in column "state" of relation "users" violates not-null constraint (SQLSTATE=23502)
`

Thanks

j2gg0s:
It should work.

Could you provide a reproducible example?

fhendrikx:
I will do that shortly.

fhendrikx:
I've attached a sample that reproduces the issue. It looks like the issue occurs on fields that don't have a default set in the database. 

So, in this example, the field "state" has a default set in the database, but field "testint" doesn't... and the code fails. If I alter the table definition to include a default, then it all works.

Table definition:

```
                        Table "public.blob"
 Column  |          Type          | Collation | Nullable | Default
---------+------------------------+-----------+----------+---------
 id            | bigint                  |             |                    |
 name      | char varying(150)  |         |                   |
 state      | integer                |             | not null      | 4
 testint    | smallint              |              | not null      |
```

I don't know if this is the desired behaviour, but seems odd to have to specify a "default" value in the bun model for stuff the database will do anyway... and I was assuming that a bun model definition for default would override the db definition..?

[main.txt](https://github.com/user-attachments/files/17594003/main.txt)

j2gg0s:
1. pgdialect will enable DeafultPlaceholder https://github.com/uptrace/bun/blob/master/dialect/pgdialect/dialect.go#L39
2. In that case, bun will use `DEFAULT` AS default https://github.com/uptrace/bun/blob/master/query_insert.go#L337

I think Bun does this assuming that your table structure matches the definition of your struct.
If that’s not the case, you can manually disable DefaultPlaceholder and use Go’s default values instead.

fhendrikx:
Thanks for your response, can you please explain how I can enable/disable a feature flag in the dialect?

j2gg0s:
[...]

PR #1070, opened 2024-11-21
feat: support disable dialect's feature

j2gg0s:
Issue: #1045

### DIFF
--- a/db.go
+++ b/db.go
@@ -35,8 +35,7 @@ func WithDiscardUnknownColumns() DBOption {
 type DB struct {
 	*sql.DB
 
-	dialect  schema.Dialect
-	features feature.Feature
+	dialect schema.Dialect
 
 	queryHooks []QueryHook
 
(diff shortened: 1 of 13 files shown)


vmihailenco:
I'd say this should be done one the dialect level, e.g.:

```go
myDialect := pgdialect.New(pgdialect.WithoutFeature(...))
```

And then the `db` will just use the features from the dialect:

```
db := bun.NewDB(sqldb, myDialect)
```

vmihailenco:
The problem with your current approach is that now we have `bun.DB` with one set of features and `dialect` with another set features. It might be that we currently don't use `dialect.features` anywhere, but it will probably create a problem in the long term.

>And i dont want to add method for every dialect.

We can try to create `BaseDialect` and extract this functionality to the `dialect` package, but TBH it looks easier to just copy/paste 2 simple methods in each package.

j2gg0s:
@vmihailenco 

Let use remove DB.features, and use only the dialect?

vmihailenco:
Looks good :+1: 

I still think that options should be named `WithFeature` and `WithoutFeature`, not `AddFeature` and `RemoveFeature`.

vmihailenco (approved):

PR #1074, opened 2024-11-22
feat: add an ordered map to remove unnecessary dependencies

Tiscs:
In a past commit (https://github.com/uptrace/bun/commit/a08c009c58bf3283ce0d39d4aff04beaeff58417), a non-tagged version of [go-ordered-map v2](https://pkg.go.dev/github.com/wk8/go-ordered-map/v2@v2.1.9-0.20240816141633-0a40785b4f41) was introduced, but only for the ordered traversal of the table and column information, and some unnecessary indirect dependencies were introduced.

This PR provides an in-package ordered map implementation so that [go-ordered-map v2](https://pkg.go.dev/github.com/wk8/go-ordered-map/v2@v2.1.9-0.20240816141633-0a40785b4f41) and its related dependencies can be removed, and returned to go 1.22, which is still widely used.


Tiscs:
> @Tiscs
> 
> Do you mind handling the updates on master?

Rebased to uptrace/bun:master now.

j2gg0s:
@vmihailenco Should we consider compatibility with at least two minor versions?

vmihailenco:
>@vmihailenco Should we consider compatibility with at least two minor go versions?

Yes, let's do it.

The old map implementation was actually compatible with Go 1.22, but I've used new Go iterators which are 1.23+ only. But the new map is also fine.

vmihailenco (approved):

PR #1080, opened 2024-11-27
fix: has many relation with driver.Valuer

thecampagnards:
Hello,
I have some trouble to use `driver.Valuer` on has many.
Here a pr, tell me if it's the wrong place
Thx !

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -279,6 +279,7 @@ func TestDB(t *testing.T) {
 		{testWithForeignKeys},
 		{testWithForeignKeysHasMany},
 		{testWithPointerForeignKeysHasMany},
+		{testWithPointerForeignKeysHasManyWithDriverValuer},
 		{testInterfaceAny},
 		{testInterfaceJSON},
 		{testScanRawMessage},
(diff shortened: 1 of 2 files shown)


Tiscs:
LGTM.

Thanks for your contribution. 

Based on the unit tests you provided, I ran some tests. Before the `indirectFieldValue` was modified, the Key value type in the User model's `baseValues` was inconsistent with the Deck model's `structKey` value type, causing `parkStruct` to return an error when matching.

I also tested the case where `User.DeskID` was a `sql.NullInt64{Valid: false}` or defined as *int64, and the code ran as expected.

I think this fix is ​​correct, @vmihailenco what do you think?

thecampagnards:
> LGTM.
> 
> Thanks for your contribution.
> 
> Based on the unit tests you provided, I ran some tests. Before the `indirectFieldValue` was modified, the Key value type in the User model's `baseValues` was inconsistent with the Deck model's `structKey` value type, causing `parkStruct` to return an error when matching.
> 
> I also tested the case where `User.DeskID` was a `sql.NullInt64{Valid: false}` or defined as *int64, and the code ran as expected.
> 
> I think this fix is ​​correct, @vmihailenco what do you think?

Thx ! Do you want me to add a `Valid: false` test ?
Thank you for checking the PR !

Issue #624, opened 2022-07-22
bug(pgdialect): `uint32` field will overflow


ionling:
https://github.com/uptrace/bun/blob/92b81c341b2908dd2dde344e20f3d7e816f63e2c/dialect/pgdialect/dialect.go#L104

If the value of a `uint32` field is `2881420146`, the final inserted value will be `-1413547150`, it's overflowed.

vmihailenco:
Postgres does not support unsigned numbers. Use `int32` or `int64` if you are worried about overflows.

ionling:
Yeah, the Postgres does not have uint32 type, but in my code I use `uint32` for avoiding the none sense type conversion.
And I think if the ORM support uint32 type, it should handle it correctly, instead of letting it overflow, and make the developer choose a right type for it. No one knows `2881420146` will overflow, if they don't check the database.

Aoang:
Another situation arises when the user specifies the type as `numeric`. This conversion behavior can lead to data anomalies. Maybe, instead of converting it ourselves, we should let the database throw an error.

j2gg0s:
@vmihailenco

We can avoid truncating here and instead let the database throw an error?

Explicit errors are always better than silent ones.

vmihailenco:
@j2gg0s such change is fine with me, but technically we don't truncate anything. All the bits are preserved.

vmihailenco:
>Explicit errors are always better than silent ones.

Agreed :+1:

PR #1086, opened 2024-12-05
fix(pgdialect): Remove unsigned integer conversion

Aoang:
PostgreSQL does not support unsigned integer types. Previously, a Go uint32 type was mapped to the database integer type. When values exceeded the range, they were converted.
Even specifying the type as numeric did not prevent this conversion. This update removes the conversion, allowing the database to handle errors appropriately.

Fixed uptrace/bun#624

### DIFF
--- a/dialect/pgdialect/dialect.go
+++ b/dialect/pgdialect/dialect.go
@@ -3,7 +3,6 @@ package pgdialect
 import (
 	"database/sql"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/uptrace/bun"
@@ -126,14 +125,6 @@ func (d *Dialect) onField(field *schema.Field) {
 
 func (d *Dialect) IdentQuote() byte {
 	return '"'
-}
-
-func (d *Dialect) AppendUint32(b []byte, n uint32) []byte {
-	return strconv.AppendInt(b, int64(int32(n)), 10)
-}
-
-func (d *Dialect) AppendUint64(b []byte, n uint64) []byte {
-	return strconv.AppendInt(b, int64(n), 10)
 }
 
 func (d *Dialect) AppendSequence(b []byte, _ *schema.Table, _ *schema.Field) []byte {


j2gg0s:
I’d like to elaborate on the specific impacts.

Assume we use uint32 in a struct in Go, Bun will create table with `Integer` in Postgres.
```go
type User struct {
        ID     int `bun:",pk"`
        Name   string
        Uint32 uint32
        Int64  int64
}
```
```sql
bun=# \d users;
                    Table "public.users"
 Column |       Type        | Collation | Nullable | Default
--------+-------------------+-----------+----------+---------
 id     | bigint            |           | not null |
 name   | character varying |           |          |
 uint32 | integer           |           |          |
 int64  | bigint            |           |          |
Indexes:
    "users_pkey" PRIMARY KEY, btree (id)
```

If we insert 1<<31+1, we got -2147483647.
But we can got it by 1<<31+1.
```go

	users := []*User{
		{ID: 1, Name: "user 1", Uint32: 1<<31 - 1},
		{ID: 2, Name: "user 2", Uint32: 1<<31 + 1},
	}
	if _, err := db.NewInsert().Model(&users).Exec(ctx); err != nil {
		panic(err)
	}

	{
		user := User{}
		if err := db.NewSelect().Model(&user).Where("uint32 = ?", uint32(1<<31+1)).Scan(ctx); err != nil {
			panic(err)
		}
		fmt.Println(user)
	}
```
```sql
[bun]  13:57:53.983   INSERT                  742µs  INSERT INTO "users" ("id", "name", "uint32", "int64") VALUES (1, 'user 1', 2147483647, 0), (2, 'user 2', -2147483647, 0)
[bun]  13:57:53.983   SELECT                  314µs  SELECT "user"."id", "user"."name", "user"."uint32", "user"."int64" FROM "users" AS "user" WHERE (uint32 = -2147483647)
```

With this PR, we will got error when insert:
```sql
[bun]  13:58:43.396   INSERT                1.541ms  INSERT INTO "users" ("id", "name", "uint32", "int64") VALUES (1, 'user 1', 2147483647, 0), (2, 'user 2', 2147483649, 0)    pgdriver.Error: ERROR: integer out of range (SQLSTATE=22003)
```

Although this may break some edge cases of runnable code, I still believe this is a good change.

PR #1088, opened 2024-12-18
fix: support scan float32 to float32

### LINKED ISSUE #1087: Bun cannot scan float32 into float32 for MySQL FLOAT columns
When using Bun with MySQL and a column of type `FLOAT`, the `database/sql` driver correctly returns a value of type `float32`. However, Bun fails to scan this value into a Go struct field of type `float32` with the following error:
My code: https://github.com/WarhoopAll/Grimoire/blob/28c38ff89bdaa980ea1618c7e8802e3f991db6ca/app/model/character.go#L23

**To Reproduce**

Create a MySQL table with a `FLOAT` column:
   ```sql
   CREATE TABLE characters (
       guid INT PRIMARY KEY,
       position_x FLOAT
   );
   INSERT INTO characters (guid, position_x) VALUES (1, -6279.93);
   ```

   ```go
type DBCharacter struct {
    PositionX float32 `bun:"position_x"`
}
```

```go
var character DBCharacter
err := bunDB.NewSelect().
    Model(&character).
    Where("guid = ?", 1).
    Scan(ctx)
if err != nil {
    log.Fatalf("Error: %v", err)
}
```

**Fast sctipt for test**
```go
package main

import (
  "context"
  "database/sql"
  "fmt"
  "log"

  _ "github.com/go-sql-driver/mysql"
  "github.com/uptrace/bun"
  "github.com/uptrace/bun/dialect/mysqldialect"
)

func main() {
  dsn := "root:root@tcp(127.0.0.1:3306)/characters"


  sqlDB, err := sql.Open("mysql", dsn)
  if err != nil {
    log.Fatalf("Failed to open database: %v", err)
  }
  defer sqlDB.Close()

  // === go-sql-driver SQL-===
  var positionX interface{}
  err = sqlDB.QueryRow("SELECT position_x FROM characters WHERE guid = ?", 1).Scan(&positionX)
  if err != nil {
    log.Fatalf("Failed to execute query (database/sql): %v", err)
  }
  fmt.Printf("Raw value (database/sql): %v, Type: %T\n", positionX, positionX)

  // === Bun  mysqldialect ===
  bunDB := bun.NewDB(sqlDB, mysqldialect.New()) 
  defer bunDB.Close()

  var positionXWithBun float32
  err = bunDB.NewRaw("SELECT position_x FROM characters WHERE guid = ?", 1).
    Scan(context.Background(), &positionXWithBun)
  if err != nil {
    log.Fatalf("Failed to execute query (Bun): %v", err)
  }
  fmt.Printf("Value from Bun (mysqldialect): %f\n", positionXWithBun)
}
```
```go
First go-sql-driver

Raw value (database/sql): -6279.927, Type: float32 - 32 type
```
```go
Second mysqldialect

2024/12/17 23:07:05 Failed to execute query (Bun): sql: Scan error on column index 0, name "position_x": bun: can't scan -6279.93 (float32) into float32
Bun cannot scan float32 even in the float32 field
```
```go
Change type 

var positionXWithBun float64

2024/12/17 23:40:32 Failed to execute query (Bun): sql: Scan error on column index 0, name "position_x": bun: can't scan -6279.93 (float32) into float64
```
okey check code 
https://github.com/uptrace/bun/blob/014b142c4ec4a0266fc7d82d1f4f765261246586/schema/scan.go#L41
reflect.Float32:       scanFloat64
there is no conversion of float32 to float64

add case
```go
func scanFloat64(dest reflect.Value, src interface{}) error {
	switch src := src.(type) {
	case nil:
		dest.SetFloat(0)
		return nil
	case float32: // Safely convert float32 to float64
		dest.SetFloat(float64(src))
		return nil
	case float64:
		dest.SetFloat(src)
		return nil
	case []byte:
		f, err := strconv.ParseFloat(string(src), 64)
		if err != nil {
			return err
		}
		dest.SetFloat(f)
		return nil
	case string:
		f, err := strconv.ParseFloat(src, 64)
		if err != nil {
			return err
		}
		dest.SetFloat(f)
		return nil
	default:
		return scanError(dest.Type(), src)
	}
}

```
work fine:
```log
Raw value (database/sql): -6279.927, Type: float32
Value from Bun (mysqldialect): -6279.930176

if now update 	var positionXWithBun float64

Raw value (database/sql): -6279.927, Type: float32
Value from Bun (mysqldialect): -6279.930176

work
```

```sh
mysql> SELECT CAST(position_x AS DOUBLE) AS precise_value FROM characters WHERE guid = 1;
+------------------+
| precise_value    |
+------------------+
| -6279.9267578125 |
+------------------+
1 row in set (0.02 sec)

mysql> SELECT position_x FROM characters WHERE guid = 1;
+------------+
| position_x |
+------------+
|   -6279.93 |
+------------+
1 row in set (0.01 sec)

mysql> SELECT VERSION();
+-----------+
| VERSION() |
+-----------+
| 8.0.40    |
+-----------+
1 row in set (0.01 sec)

mysql> DESCRIBE characters position_x;
+------------+-------+------+-----+---------+-------+
| Field      | Type  | Null | Key | Default | Extra |
+------------+-------+------+-----+---------+-------+
| position_x | float | NO   |     | 0       |       |
+------------+-------+------+-----+---------+-------+
1 row in set (0.03 sec)

go list -m github.com/go-sql-driver/mysql
github.com/go-sql-driver/mysql v1.8.1

go list -m github.com/uptrace/bun/dialect/mysqldialect
github.com/uptrace/bun/dialect/mysqldialect v1.2.6
```

Old issue
https://github.com/uptrace/bun/issues/993

j2gg0s:
Close #1087

### DIFF
--- a/schema/scan.go
+++ b/schema/scan.go
@@ -38,8 +38,8 @@ func init() {
 		reflect.Uint32:        scanUint64,
 		reflect.Uint64:        scanUint64,
 		reflect.Uintptr:       scanUint64,
-		reflect.Float32:       scanFloat64,
-		reflect.Float64:       scanFloat64,
+		reflect.Float32:       scanFloat,
+		reflect.Float64:       scanFloat,
 		reflect.Complex64:     nil,
 		reflect.Complex128:    nil,
 		reflect.Array:         nil,
@@ -214,10 +214,13 @@ func scanUint64(dest reflect.Value, src interface{}) error {
 	}
 }
 
-func scanFloat64(dest reflect.Value, src interface{}) error {
+func scanFloat(dest reflect.Value, src interface{}) error {
 	switch src := src.(type) {
 	case nil:
 		dest.SetFloat(0)
+		return nil
+	case float32:
+		dest.SetFloat(float64(src))
 		return nil
 	case float64:
 		dest.SetFloat(src)


j2gg0s:
@vmihailenco 

Do we have any specific reason not to do this? I suspect this might be a bug.

Re3os:
@j2gg0s 
Why duplicate the code when we can just use:
  ```go
  case float32:
      dest.SetFloat(float64(src))
      return nil
    ```
    Or bring the function to common values:
    ```go
    func scanFloat(dest reflect.Value, src interface{}) error {
    ```
    It's wrong to duplicate code, I may be wrong?

j2gg0s:
> @j2gg0s Why duplicate the code when we can just use:
> 
> ```go
> case float32:
>     dest.SetFloat(float64(src))
>     return nil
> ```
> 
> ```
> Or bring the function to common values:
> 
> func scanFloat(dest reflect.Value, src interface{}) error
> 
> It's wrong to duplicate code, I may be wrong?
> ```

@Re3os 
Currently, we dont support scan float32 to float64.
I dont want to break this.

To me, there is no difference in merit between adding an if condition and duplicating code.

schema/scan.go
```diff
@@ -214,6 +214,36 @@ func scanUint64(dest reflect.Value, src interface{}) error {
...
+		dest.SetFloat(0)
+		return nil
+	case float32:
+		dest.SetFloat(float64(src))
+		return nil
+	case float64:
+		dest.SetFloat(src)
+		return nil
+	case []byte:
+		f, err := strconv.ParseFloat(internal.String(src), 64)
```
vmihailenco, line 229: Probably should be 64 -> 32

vmihailenco:
>Currently, we dont support scan float32 to float64.

But why? You allow to scan float64 into float32 which results in precision loss, but want to forbid scanning float32 to float64 which is fine.

I agree with @Re3os that this could be refactored into a more general function:

````
func scanFloat32(dest reflect.Value, src interface{}) error {
    return scanFloat(dest, src, 32)
}

func scanFloat64(dest reflect.Value, src interface{}) error {
    return scanFloat(dest, src, 64)
}

func scanFloat64(dest reflect.Value, src interface{}, bits int) error {
}
```

>Do we have any specific reason not to do this? I suspect this might be a bug.

AFAIR no I think I assumed that database/sql never returns float32.

Re3os:
@j2gg0s 
```go
func scanFloat(dest reflect.Value, src interface{}, bits int) error {
    var parsedFloat float64 

    switch src := src.(type) {
    case nil:
        parsedFloat = 0
    case float32:
        parsedFloat = float64(src)
    case float64:
        parsedFloat = src
    case []byte:
        f, err := strconv.ParseFloat(string(src), bits)
        if err != nil {
            return err
        }
        parsedFloat = f
    case string:
        f, err := strconv.ParseFloat(src, bits)
        if err != nil {
            return err
        }
        parsedFloat = f
    default:
        return scanError(dest.Type(), src)
    }

    if bits == 32 && dest.Kind() == reflect.Float32 {
        dest.SetFloat(float64(float32(parsedFloat))) 
    } else if bits == 64 && dest.Kind() == reflect.Float64 {
        dest.SetFloat(parsedFloat)
    } else {
        return scanError(dest.Type(), src)
    }

    return nil
}
```

and 

```go
func scanFloat32(dest reflect.Value, src interface{}) error {
    return scanFloat(dest, src, 32)
}

func scanFloat64(dest reflect.Value, src interface{}) error {
    return scanFloat(dest, src, 64)
}

```

j2gg0s:
Current logic in master:
1. support scan float64 to float64
2. support scan float64 to float32
3. dont support scan float32 to float64
4. dont support scan float32 to float32

I think:
1 must be correct.
2 might be incorrect because it could lose precision
3 might be supported, but it is not mandatory.
4 must be incorrect

Due to personal habits, I am unwilling to modify any existing logic unless I have confirmed how everyone uses it, or it is obviously incorrect.

However, I think it is also acceptable to directly support 3 & 4 now.

schema/scan.go
j2gg0s: I made a revision; perhaps we don’t need to differentiate here anymore.

Re3os:
@vmihailenco any news?

PR #1089, opened 2024-12-19
chore(internal): utilize methods from the internal package

Aoang:
- Remove redundant code for improved maintainability
- These sections will be inlined during compilation


driver/pgdriver/safe.go
```diff
@@ -1,11 +0,0 @@
-// +build appengine
```
j2gg0s, line 1: @vmihailenco @Aoang 

Do we need to implement any special logic for Google App Engine?

Aoang: The internal package has compatible.

https://github.com/uptrace/bun/blob/v1.2.6/internal/safe.go

https://github.com/uptrace/bun/blob/v1.2.6/internal/unsafe.go

Issue #554, opened 2022-06-02
How to add custom where clause on Relation?


imraan-go:
Is there any way to add additional condition when using Relation ?
```go
q := repo.db.NewSelect().Model(&product)
q.Relation("Stock", func(q *bun.SelectQuery) *bun.SelectQuery {
			return q.Where("branch_id = ?", filter.BranchId)
		})
```

This above code should produce   something like:
```sql
SELECT * from products as product
LEFT JOIN "stock" AS "stock" ON ( "stock"."product_id" = "product"."id" ) 
	AND stock.branch_id = 5
```
The Where condition should be inside the left join clause. But currently it produces something like:

```sql
SELECT * from products as product
LEFT JOIN "stock" AS "stock" ON ( "stock"."product_id" = "product"."id" ) 
WHERE branch_id = 5
```

Therefore left join does not work. I can do the same with `Join()` and `JoinOn()` but then Bun does not populate struct for some reason.

vmihailenco:
How `q.Where("branch_id = ?", filter.BranchId)` is better than using `q.JoinOn("branch_id = ?", filter.BranchId)`?

fiftin:
If we implement this we broke current behaviour. @vmihailenco is this ok?

vmihailenco:
I think we should make this work for has-one and belongs-to relations:

```go
q.Relation("Stock", func(q *bun.SelectQuery) *bun.SelectQuery {
			return q.JoinOn("branch_id = ?", filter.BranchId)
		})
```

fiftin:
I mean replacing this:

```sql
SELECT * from products as product
LEFT JOIN "stock" AS "stock" ON ( "stock"."product_id" = "product"."id" ) 
WHERE branch_id = 5
```

to this:

```sql
SELECT * from products as product
LEFT JOIN "stock" AS "stock" ON "stock"."product_id" = "product"."id" AND stock.branch_id = 5
```

Some users may add filters to global WHERE section from callback. After this change it will no longer be possible.

fiftin:
@vmihailenco 

Should it be related to https://github.com/uptrace/bun/issues/557? Because `"branch_id = ?"` can return conflict error.

imraan-go:
> .Where("branch_id = ?", filter.BranchId)

I can't seem to make q.JoinOn("branch_id = ?", filter.BranchId) work. Do you have any example that will additional conditional Join statemment?   I've tried custom join and joinOn combination like this
```golang

[...]

PR #1097, opened 2025-01-02
Add RelationWithOpts to SelectQuery for Customizing JOIN ON Clause

takaaa220:
## Description
Closes https://github.com/uptrace/bun/issues/554.

This PR introduces a new method, RelationWithOpts, to the SelectQuery interface. This method allows users to customize the JOIN ON clause when defining relationships between models (e.g., has-one, has-many).

## Key Points of Discussion

Current Behavior: The existing Relation method allows applying filters to the joined table's SELECT query but does not modify the JOIN ON clause. For example(it is on https://github.com/uptrace/bun/issues/554):

```go
q := repo.db.NewSelect().Model(&product)
q.Relation("Stock", func(q *bun.SelectQuery) *bun.SelectQuery {
    return q.Where("branch_id = ?", filter.BranchId)
})
```

- I believe it's essential to preserve this behavior to maintain backward compatibility and separation of concerns (filtering the joined table vs. customizing JOIN ON).
- Proposed Changes: The RelationWithOpts method provides an additional option to customize the JOIN ON clause while keeping the existing functionality of Relation.

## Example Usage

Here is a quick example of how RelationWithOpts can be utilized:

```go
q := repo.db.NewSelect().Model(&product)
q.RelationWithOpts("Stock", &bun.RelationOpts{
	AdditionalJoinOnConditions: []schema.QueryWithArgs{
		{
			Query: "stock.quantity = ?",
			Args:  []any{1},
		},
	},
})
```

I add this feature but actually i don't like this feature itself and the interface very much (especially, i don't like to directly face `[]schema.QueryWithArgs`). Feedback and suggestions for improving the interface and functionality are welcome. Let me know your thoughts! (If this feature is deemed unnecessary, please feel free to close this pull request without hesitation.)

### DIFF
--- a/internal/dbtest/orm_test.go
+++ b/internal/dbtest/orm_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/uptrace/bun/dbfixture"
 	"github.com/uptrace/bun/dialect"
 	"github.com/uptrace/bun/dialect/feature"
+	"github.com/uptrace/bun/schema"
 )
 
 func TestORM(t *testing.T) {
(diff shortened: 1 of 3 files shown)


j2gg0s:
I will release v1.2.9 tonight.

PR #1098, opened 2025-01-02
feat: sort fields by struct

### LINKED ISSUE #1095: Column Ordering Changed after v1.2.0 for Embedded Structs
### Description
After upgrading to bun v1.2.0, the column ordering in generated SQL tables has changed when using embedded structs. Previously, columns were created in the order they were defined in the struct (including embedded structs). However, in v1.2.0, the columns from embedded structs appear after the directly defined columns, regardless of their position in the struct definition.

### Reproduction
```Go
// 20210505110026_foo.go 
package main
type BaseModel struct {
	ID        int
	CreatedAt time.Time `bun:",default:current_timestamp"`
}
type FooModel struct {
	BaseModel

	Name string
}

func init() {
	Migrations.MustRegister(func(ctx context.Context, db *bun.DB) error {
		_, err := db.NewCreateTable().Model((*FooModel)(nil)).Exec(ctx)
		return err
	}, func(ctx context.Context, db *bun.DB) error {
		_, err := db.NewDropTable().Model((*FooModel)(nil)).Exec(ctx)
		return err
	})
}
```

```Go
// main.go
package main
import (
	"context"
	"database/sql"

	"github.com/uptrace/bun"
	"github.com/uptrace/bun/migrate"
)

var Migrations = migrate.NewMigrations()

func main() {
	sqldb, err := sql.Open(sqliteshim.ShimName, "file:test.s3db?cache=shared")
	if err != nil {
		panic(err)
	}

	if err := Migrations.DiscoverCaller(); err != nil {
		panic(err)
	}
	migrator := migrate.NewMigrator(db, Migrations)

	ctx := context.Background()
	err = migrator.Init(ctx)
	if err != nil {
		panic(err)
	}

	_, err = migrator.Migrate(ctx)
	if err != nil {
		panic(err)
	}
}

```

### Expected Behavior
The generated table should have columns in the order they are defined in the struct hierarchy:
1. `id`
2. `created_at`
3. `name`

### Actual Behavior
After v1.2.0, the generated table has columns in this order:
1. `name`
2. `id`
3. `created_at`

The columns from the embedded `BaseModel` struct are placed after the directly defined `Name` column, which breaks the expected structural order.

### Version Information
- Before: Working as expected in versions prior to v1.2.0
- After: Column ordering changed in v1.2.0 and above

### Impact
This change affects systems that rely on consistent column ordering, particularly in scenarios where:
- Database tools or ORMs assume a specific column order
- Migration scripts depend on predictable column positioning
- Legacy system integrations expect a particular column sequence

### Additional Notes
The change in column ordering appears to be related to how bun v1.2.0 handles embedded struct fields during table creation. While this doesn't affect functionality, it represents a breaking change in terms of database schema generation.

j2gg0s:
Close #1095

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -250,6 +251,30 @@ func (t *Table) processFields(typ reflect.Type) {
 			t.addUnique(subfield, embfield.prefix, v)
 		}
 	}
+
+	if len(embedded) > 0 {
+		// https://github.com/uptrace/bun/issues/1095
+		// < v1.2, all fields follow the order corresponding to the struct
+		// >= v1.2, < v1.2.8, fields of nested structs have been moved to the end.
+		// >= v1.2.8, The default behavior remains the same as initially,
+		sortFieldsByStruct(t.allFields)
+		sortFieldsByStruct(t.Fields)
+		sortFieldsByStruct(t.PKs)
+		sortFieldsByStruct(t.DataFields)
+	}
+}
+
+func sortFieldsByStruct(fields []*Field) {
+	sort.Slice(fields, func(i, j int) bool {
+		left, right := fields[i], fields[j]
+		for k := 0; k < len(left.Index) && k < len(right.Index); k++ {
+			if left.Index[k] != right.Index[k] {
+				return left.Index[k] < right.Index[k]
+			}
+		}
+		// NOTE: should not reach
+		return true
+	})
 }
 
 func (t *Table) addUnique(field *Field, prefix string, tagOptions []string) {


schema/table.go
```diff
@@ -24,6 +25,13 @@ const (
 var (
 	baseModelType      = reflect.TypeFor[BaseModel]()
 	tableNameInflector = inflection.Plural
+
+	// https://github.com/uptrace/bun/issues/1095
+	// < v1.2, all fields follow the order corresponding to the struct
+	// >= v1.2, < v1.2.8, fields of nested structs have been moved to the end.
+	// >= v1.2.8, The default behavior remains the same as initially,
+	// but we allow users to control this behavior through this parameter.
+	DisableSortFieldsByStruct = false
```
vmihailenco, line 34: I'd prefer that we don't introduce this option and instead always preserve the order of struct fields.

schema/table.go
```diff
@@ -250,6 +258,26 @@ func (t *Table) processFields(typ reflect.Type) {
 			t.addUnique(subfield, embfield.prefix, v)
 		}
 	}
+
+	if len(embedded) > 0 && !DisableSortFieldsByStruct {
```
vmihailenco, line 262: @j2gg0s let's remove DisableSortFieldsByStruct and always sort the fields. Are there any undesired consequences?

PR #1099, opened 2025-01-03
fix: return error when use dest with has-many/many-to-many.

### LINKED ISSUE #606: Scan() does not populate has-many relationships
In context of the ['has-many' example](https://github.com/uptrace/bun/tree/master/example/rel-has-many), if you do:

    users := make([]User, 0)

    db.NewSelect().
        Model((*User)(nil)).
        Relation("Profiles").
        Scan(ctx, &users)

`users[k].Profiles` will always be `nil`, regardless of the related Profiles they might have.

I don't think this is by design, as the [documentation](https://bun.uptrace.dev/guide/queries.html#scan-and-exec) states that you can pass a struct as a 'destination', and no mention about the above behaviour.

j2gg0s:
Close #606

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -838,6 +838,14 @@ func (q *SelectQuery) scanResult(ctx context.Context, dest ...interface{}) (sql.
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


j2gg0s:
Is there a way to fix it instead of returning an error?

PR #1110, opened 2025-01-22
feat(pgdialect): allow to convert uint to int

vmihailenco:
This allow to achieve the behavior removed in https://github.com/uptrace/bun/pull/1086/files via an option


j2gg0s:
Can you introduce a user case to help us understand better?

Issue #877, opened 2023-08-04
Invalid Postgres Insert Syntax for Slices of String Pointers


pc-bob:
I have a struct defined like this:

```golang
type Location struct {
	ID string `bun:"type:uuid,pk"`

	MyStrings []string `bun:"type:text[],notnull"`

	MyOptionalStrings []*string `bun:"type:text[]"`

}
```

The postgres statement generated when trying to insert a value for this model does not correctly quote the string values within the array:

```
INSERT INTO "locations" AS "location" ("id", "my_strings", "my_optional_strings") VALUES ('C99C8B07-B153-4C1C-8CD9-907FC06FD46D', '{"foo","bar"}', '{'baz'}')
```

This results in an error from postgres:

```
*pq.Error: pq: syntax error at or near "baz"
```

The non-nullable string slice works correctly, but the slice of string pointers generates `'{'baz'}'` when it should generate `{"baz"}'` to be valid.

PR #1111, opened 2025-01-22
fix(pgdialect): postgres syntax errors for pointers and slices #877

rfarrjr:
Addresses issue identified in #877.

New tests added for pointers to slice and slice of pointer elements including arrays of JSONB columns.

### DIFF
--- a/dialect/pgdialect/array.go
+++ b/dialect/pgdialect/array.go
@@ -142,6 +142,10 @@ func (d *Dialect) arrayElemAppender(typ reflect.Type) schema.AppenderFunc {
 	if typ.Implements(driverValuerType) {
 		return arrayAppendDriverValue
 	}
+	if typ == timeType {
+		return appendTimeElemValue
+	}
+
 	switch typ.Kind() {
 	case reflect.String:
 		return appendStringElemValue
@@ -153,6 +157,14 @@ func (d *Dialect) arrayElemAppender(typ reflect.Type) schema.AppenderFunc {
 		return schema.PtrAppender(d.arrayElemAppender(typ.Elem()))
 	}
 	return schema.Appender(d, typ)
+}
+
+func appendTimeElemValue(fmter schema.Formatter, b []byte, v reflect.Value) []byte {
+	ts := v.Convert(timeType).Interface().(time.Time)
+
+	b = append(b, '"')
+	b = appendTime(b, ts)
+	return append(b, '"')
 }
 
 func appendStringElemValue(fmter schema.Formatter, b []byte, v reflect.Value) []byte {
--- a/dialect/pgdialect/array.go
+++ b/dialect/pgdialect/array.go
@@ -149,6 +149,8 @@ func (d *Dialect) arrayElemAppender(typ reflect.Type) schema.AppenderFunc {
 		if typ.Elem().Kind() == reflect.Uint8 {
 			return appendBytesElemValue
 		}
+	case reflect.Ptr:
+		return schema.PtrAppender(d.arrayElemAppender(typ.Elem()))
 	}
 	return schema.Appender(d, typ)
 }
--- a/dialect/pgdialect/array_parser.go
+++ b/dialect/pgdialect/array_parser.go
@@ -18,6 +18,11 @@ type arrayParser struct {
 func newArrayParser(b []byte) *arrayParser {
 	p := new(arrayParser)
 
+	if b[0] == 'n' {
+		p.p.Reset(nil)
+		return p
+	}
+
 	if len(b) < 2 || (b[0] != '{' && b[0] != '[') || (b[len(b)-1] != '}' && b[len(b)-1] != ']') {
 		p.err = fmt.Errorf("pgdialect: can't parse array: %q", b)
 		return p
--- a/dialect/pgdialect/array_parser.go
+++ b/dialect/pgdialect/array_parser.go
@@ -11,15 +11,18 @@ type arrayParser struct {
 
 	elem []byte
 	err  error
+
+	isJson bool
 }
 
 func newArrayParser(b []byte) *arrayParser {
 	p := new(arrayParser)
 
-	if len(b) < 2 || b[0] != '{' || b[len(b)-1] != '}' {
+	if len(b) < 2 || (b[0] != '{' && b[0] != '[') || (b[len(b)-1] != '}' && b[len(b)-1] != ']') {
 		p.err = fmt.Errorf("pgdialect: can't parse array: %q", b)
 		return p
 	}
+	p.isJson = b[0] == '['
 
 	p.p.Reset(b[1 : len(b)-1])
 	return p
@@ -51,7 +54,7 @@ func (p *arrayParser) readNext() error {
 	}
 
 	switch ch {
-	case '}':
+	case '}', ']':
 		return io.EOF
 	case '"':
 		b, err := p.p.ReadSubstring(ch)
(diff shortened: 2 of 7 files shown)


internal/dbtest/pg_test.go
```diff
@@ -877,3 +887,51 @@ func TestPostgresMultiRange(t *testing.T) {
...
+
+func TestPostgresJSONB(t *testing.T) {
+	type Item struct {
+		Name string `json:"name"`
+	}
+	type Model struct {
+		ID      int64   `bun:",pk,autoincrement"`
+		Item    Item    `bun:",type:jsonb"`
+		ItemPtr *Item   `bun:",type:jsonb"`
+		Items   []Item  `bun:",type:jsonb,array"`
```
j2gg0s, line 899: What does it mean to use both jsonb and array together?
 I think PostgreSQL will handle the JSONB type, and we only need to consider JSONB when inserting and reading data?

j2gg0s:
Thanks your PR, it seems like it could help us a lot.

I have a strange example, not sure if it’s appropriate.
The model:
```go
type Location struct {
	ID string `bun:"type:uuid,pk"`

	// Users []User `bun:"type:jsonb,array"`
	UserIDs []UserID `bun:"type:text[]"`
}

type UserID struct {
	ID string
}
```
Run with output:
```shell
[bun]  12:21:27.085   DROP TABLE           49.112ms  DROP TABLE IF EXISTS "locations" CASCADE
[bun]  12:21:27.102   CREATE TABLE         16.262ms  CREATE TABLE "locations" ("id" uuid NOT NULL, "user_i_ds" text[], PRIMARY KEY ("id"))
[bun]  12:21:27.109   INSERT                7.047ms  INSERT INTO "locations" ("id", "user_i_ds") VALUES ('a2cd66d9-c5c6-4b06-a0a9-da2b4ed85960', '["foo","bar"]')       pgdriver.Error: ERROR: malformed array literal: "["foo","bar"]" (SQLSTATE=22P02)
panic: ERROR: malformed array literal: "["foo","bar"]" (SQLSTATE=22P02)
```

dialect/pgdialect/array.go
```diff
@@ -121,18 +125,33 @@ func (d *Dialect) arrayAppender(typ reflect.Type) schema.AppenderFunc {
...
+		if kind == reflect.Slice {
+			elemType := v.Type().Elem()
+
+			if !elemType.Implements(driverValuerType) {
+				for elemType.Kind() == reflect.Ptr {
+					elemType = elemType.Elem()
+				}
+
+				if elemType.Kind() == reflect.Struct {
+					brackets = "[]"
```
j2gg0s, line 137: Is this meant to handle JSONB? But why would the JSONB type end up being processed here?

Aoang:
I believe this is intended to enhance structural support for "text[]", but the underlying logic appears to be somewhat perplexing. Perhaps I am not fully grasping it yet.

It may be beneficial for you to review the following link: https://github.com/FlowerLab/sorbifolia/blob/main/bunpgd/datatype/typ_array.go to determine if it aligns with your objectives.

internal/dbtest/pg_test.go
rfarrjr: Yes.   PostgreSQL doesn't actually support a JSONB array directly, but rather includes several functions (`jsonb_array_elements`) for working with a JSONB arrays on a JSONB column.   Can read more [here](https://hevodata.com/learn/query-jsonb-array-of-objects/).

dialect/pgdialect/array.go
rfarrjr: These are needed to format a JSON array since the underlying column is just JSONB.  

I think I need to add a check here to ignore this if the type is no JSONB or if the elemType implement query appender to handle your example above correctly.  Will add that test and fix.

rfarrjr:
This PR addresses a couple of things:

1) Handle array of structs mapping to a JSONB column, which does enhance the option of serializing/deserializing structs into a `text[]` type.

2) ,array sql type detection doesn't handle arrays of pointers with specifying the type.  i.e.  

```golang
type Person {
   Names []*string `bun:",array"`  
}
```

3) Property quoting strings in an array when using `[]*string`

j2gg0s:
In my understanding, this PR aims to accomplish two things:
1. Fix bug when Go type is `[]*string` and the PostgreSQL's type is `text[]`
2. Support append&scan, when Go type is `[]struct` and type PostgreSQL's type is `array`

For 1, this should be fixed, and we can also handle other basic types along the way.

For 2, I don’t see the necessity to support this feature for now. 
Most of the time, we use text or JSONB in PostgreSQL for storage.

rfarrjr:
This actually got me thinking, so I tried the following and it is functional.  Could be worth updating the documentation to show how to handle arrays when using jsonb.

```graphql
type Model struct {
  ID        int64    `bun:",pk,autoincrement"`
  Items     []Item   `bun:",type:jsonb"`
  ItemsP    []*Item  `bun:",type:jsonb"`
}
```

I will back out the enhancement changes and just leave the fixes

rfarrjr:
@j2gg0s I believe this addresses all of the edge cases we have

dialect/pgdialect/array.go
```diff
@@ -149,6 +149,8 @@ func (d *Dialect) arrayElemAppender(typ reflect.Type) schema.AppenderFunc {
 		if typ.Elem().Kind() == reflect.Uint8 {
 			return appendBytesElemValue
 		}
+	case reflect.Ptr:
+		return schema.PtrAppender(d.arrayElemAppender(typ.Elem()))
```
j2gg0s, line 153: A very concise implementation, but timestamp may need to be handled separately as they should also be considered a basic type.

dialect/pgdialect/array_parser.go
```diff
@@ -11,15 +11,23 @@ type arrayParser struct {
...
+	if b[0] == 'n' {
+		p.p.Reset(nil)
+		return p
+	}
+
+	if len(b) < 2 || (b[0] != '{' && b[0] != '[') || (b[len(b)-1] != '}' && b[len(b)-1] != ']') {
 		p.err = fmt.Errorf("pgdialect: can't parse array: %q", b)
 		return p
 	}
+	p.isJson = b[0] == '['
```
j2gg0s, line 30: Is this to support scan text/varchar into []T?
I think that most users would directly use JSONB/Array.

j2gg0s (commented):
Thank you very much for your work.
I’m sorry for being away for a while, which caused my delayed response until today.

I don’t think we need to support scanning text/varchar into an []T.
In most cases, JSONB/Array should be used to store []T.

We should consider timestamp as basic type, and support scan []*time.Time to timestamptz[].
We need extra code to handle this case.

dialect/pgdialect/array_parser.go
rfarrjr: This is to support scanning a JSONB/Array into a []T.

dialect/pgdialect/array.go
rfarrjr: This should handle timestamps: https://github.com/uptrace/bun/pull/1111/commits/4c4e12aa7f27cf49189427da5104afb436af4348

rfarrjr:
@j2gg0s thank you for the feedback.  

>I don’t think we need to support scanning text/varchar into an []T.
Just for context, I added this test case to assert that this change did not break existing functionality.

dialect/pgdialect/array_parser.go
j2gg0s: - If PG's type is Array, b[0] is always '{'
- If PG's type is JSONB, `scanJSON` will take effect, [schema/scan.go#L51](https://github.com/uptrace/bun/blob/master/schema/scan.go#L51)

So, what is being handled here is the case where the PG's type is TEXT and the Go type is []any. 

I hope I haven’t misunderstood.
If that’s the case, I think we should encourage users to use JSONB/Array instead of TEXT.

So, can we remove the related parsing logic?

rfarrjr: The case this is meant to handle is when trying to scan a JSONB column directly into a value vs the model.

```golang	
        var items []Item
	err = db.NewSelect().Model((*Model)(nil)).
		Column("items").
		Scan(ctx, pgdialect.Array(&items))
```

`pgdialect.Array` uses the arrayScanner, but JSONB arrays use [] vs {}.

Thinking about alternatives to this option and implementation:

```golang	
        var items []Item
	err = db.NewSelect().Model((*Model)(nil)).
		Column("items").
		Scan(ctx, &items)
```

Doesn't work because it assumes that the items array should have a column called "items" not be the item.  

Option 2: create anohter pgdialect wrapper for JSON values.  So maybe

```golang	
        var items []Item
	err = db.NewSelect().Model((*Model)(nil)).
		Column("items").
		Scan(ctx, pgdialect.JSON(&items))
```

An that wrapper can setup the appropriate scanner to be used.  

Option 3: don't support this.

PR #1122, opened 2025-02-10
feat: allow setting a query comment through a context value

kylemcc:
This is an enhancement to recently merged comment functionality. This implementation preserves the current functionality, but allows a comment to be set at the time a query is executed.

This can be used like so:

```go
qctx := bun.ContextWithComment(ctx, "traceid:"+traceID)
err := db.NewSelect().
    Model(&things).
    Where("type = ?", t).
    Scan(qctx)
```

Which, like the `Comment()` API will produce:

```sql
/* traceid:f2e006f2-8494-4ce6-947e-2c77f4c33dc8 */ SELECT * FROM thing WHERE type = 'whatever'
```

I should note that in this implementation, if a comment is set through a context, it takes precedence over one set on a query using the `Comment()` API. If the opposite precedence makes more sense, I'm happy to change this.

I know there was some discussion about supporting multiple comments - I'm happy to add some support for that if it's desirable. I think there's good reason to support both mechanisms and have them play nicely together, but maybe that should be discussed separately. For example, the idea of adding a `.Comment("descriptive query name")` to individual queries, but then being able to combine that with a context that provides a trace ID so that all queries made as part of a given request can be identified. In this case, I would propose that any comment set on the context appear first in the query followed by any set using the `Comment()` method.

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,6 +1,7 @@
 package bun
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"strings"
@@ -85,4 +86,27 @@ func appendComment(b []byte, name string) []byte {
 	name = strings.ReplaceAll(name, `/*`, `/\*`)
 	name = strings.ReplaceAll(name, `*/`, `*\/`)
 	return append(b, fmt.Sprintf("/* %s */ ", name)...)
+}
+
+// queryCommentCtxKey is a context key for setting a query comment on a context instead of calling the Comment("...") API directly
+type queryCommentCtxKey struct{}
+
+// WithComment returns a context that includes a comment that may be included in a query for debugging
+//
+// If a context with an attached query is used, a comment set by the Comment("...") API will be overwritten.
+func WithComment(ctx context.Context, comment string) context.Context {
+	return context.WithValue(ctx, queryCommentCtxKey{}, comment)
+}
+
+// commenter describes the Comment interface implemented by all of the query types
+type commenter[T any] interface {
+	Comment(string) T
+}
+
+// setCommentFromContext sets the comment on the given query from the supplied context if one is set using the Comment(...) method.
+func setCommentFromContext[T any](ctx context.Context, q commenter[T]) {
+	s, _ := ctx.Value(queryCommentCtxKey{}).(string)
+	if s != "" {
+		q.Comment(s)
+	}
 }
--- a/query_column_add.go
+++ b/query_column_add.go
@@ -137,6 +137,9 @@ func (q *AddColumnQuery) Exec(ctx context.Context, dest ...interface{}) (sql.Res
 		return nil, feature.NewNotSupportError(feature.AlterColumnExists)
 	}
 
+	// if a comment is propagated via the context, use it
+	setCommentFromContext(ctx, q)
+
 	queryBytes, err := q.AppendQuery(q.db.fmter, q.db.makeQueryBytes())
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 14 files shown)


j2gg0s:
Passing comments through context is a good idea, thank you for your contribution.

1. If both the Context and Comment methods specify a comment, wouldn’t it make more sense to merge them?
2. We can implement the logic in the [beforeQuery](https://github.com/uptrace/bun/blob/master/hook.go#L55) hook method, right? That way, the intrusion into the code will be minimal.

kylemcc:
>     1. If both the Context and Comment methods specify a comment, wouldn’t it make more sense to merge them?
I wouldn't be opposed to merging them - I think there is definitely value there. If I remember correctly, one of the other discussions raised some questions about comment order if multiple comments are specified. So, I think we just need to decide what the proper behavior should be.

>     2. We can implement the logic in the [beforeQuery](https://github.com/uptrace/bun/blob/master/hook.go#L55) hook method, right? That way, the intrusion into the code will be minimal.

Yes, I think that would be a reasonable way to refactor and DRY up this addition and the existing implementation. Happy to take a pass at that if you'd like.

j2gg0s:
@kylemcc 

I respect your opinion on merging or replacing. 
I don’t have much understanding of the use cases for this method.

kylemcc:
> > ```
> > 2. We can implement the logic in the [beforeQuery](https://github.com/uptrace/bun/blob/master/hook.go#L55) hook method, right? That way, the intrusion into the code will be minimal.
> > ```
> 
> Yes, I think that would be a reasonable way to refactor and DRY up this addition and the existing implementation. Happy to take a pass at that if you'd like.

I finally had some time to look into this, and it turns out to be a bit more complicated than I realized. the `beforeQuery` hook doesn't currently have the ability to modify the query. The query is built/formatted before being sent to that hook (see [here](https://github.com/uptrace/bun/blob/187743b1e743755cd57a9cc11e7f2f9cea0a7dcd/query_select.go#L799-L802) or [here](https://github.com/uptrace/bun/blob/187743b1e743755cd57a9cc11e7f2f9cea0a7dcd/query_base.go#L609), for example). (I actually tried to implement this as a hook to make it an optional feature, which led me to that discovery). We could make it possible to modify the query in that hook, but in my opinion that would make the code harder to follow.

> @kylemcc
> 
> I respect your opinion on merging or replacing. I don’t have much understanding of the use cases for this method.

Given the above, I think my suggestion would be to stick with the current implementation where the context comment replaces the comment set on the query/provides an alternate way to set a comment. We can always add support for multiple comments or merging the two if there's demand, but it doesn't seem like anyone is asking for that yet.

j2gg0s:
I’m sorry for not checking the logic of beforeQuery, which ended up wasting your time.

util.go
```diff
@@ -86,3 +87,26 @@ func appendComment(b []byte, name string) []byte {
...
 	return append(b, fmt.Sprintf("/* %s */ ", name)...)
 }
+
+// queryCommentCtxKey is a context key for setting a query comment on a context instead of calling the Comment("...") API directly
+type queryCommentCtxKey struct{}
+
+// ContextWithComment returns a context that includes a comment that may be included in a query for debugging
+//
+// If a context with an attached query is used, a comment set by the Comment("...") API will be overwritten.
+func ContextWithComment(ctx context.Context, comment string) context.Context {
```
j2gg0s, line 97: Would you mind renaming it to WithComment to keep it consistent with WithCancel / WithTimeout and similar naming conventions?

kylemcc: Done!

kylemcc:
> I’m sorry for not checking the logic of beforeQuery, which ended up wasting your time.

No problem :)

PR #1124, opened 2025-02-13
chore(ci): Update Go version matrix in build workflow

Aoang:
Update the Go version matrix in the build workflow to include Go 1.24. This ensures that the project is tested against the latest supported Go versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go-version: [1.23.x, 1.24.x]
+        go-version: [1.22.x, 1.23.x, 1.24.x]
 
     services:
       postgres:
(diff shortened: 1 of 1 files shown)


j2gg0s:
Although the Go team officially supports only the latest two versions, 
I feel that the most widely used version is <= v1.22.

So, let’s support the latest three versions instead, especially right after the release of v1.24.

Tiscs:
I also think that support for go 1.22 should be retained.

Aoang:
Thanks to the compatibility of Go, the impact of updating Go versions on Bun is minimal.

https://github.com/uptrace/bun/blob/605cd572ad980e9e2580d55ae623c5720feaf601/go.mod#L3

The last supported version should be determined by the version specified in ⁠go.mod. Here, the change is simply that CI testing is no longer performed on this version.

Whether running tests on three versions will slow down and bulky the Pull Request process. If this isn't a concern, please reply.

j2gg0s:
If the minimum supported version is v1.22, I believe keeping it in the CI is a good idea.
Although Go version updates introduce almost zero incompatibility issues,

Intuitively, CI runs in parallel, so adding v1.24 would consume negligible resources and wouldn’t increase our waiting time.

Issue #1115, opened 2025-01-28
`rel:has-one` on a relationship should not have foreign key creation


K1DV5:
I was writing a function to generate the create table statements with the foreign keys and it was working fine...until I found out that I had a `has-one` relationship.

I was under the impression that `has-one` was meant to be used in place of `has-many` to deal with a one-to-one relationship. Example copied from the docs:

```go
// Profile belongs to User.
type Profile struct {
	ID     int64 `bun:",pk"`
	Lang   string
	UserID int64
}

type User struct {
	ID      int64 `bun:",pk"`
	Name    string
	Profile *Profile `bun:"rel:has-one,join:id=user_id"`
}
```
It says right there that the Profile belongs to the user and that is what should refer to the User not the other way around. But currently, when trying to use the `bun.CreateTableQuery.WithForeignKeys()`, it is creating a foreign key like `FOREIGN KEY ("id") REFERENCES "profiles" ("user_id")` on the *User* table.

Also, as I have a `belongs-to` relationship on the Profile model, it is correctly creating the foreign key to reference the User table. The issue is that the other one should not be created.

j2gg0s:
I’m not familiar with the code related to migration.
I need to find some free time to go through the overall logic.

😂

j2gg0s:
The example in the documentation doesn’t seem quite appropriate. 

If it’s a has-one relationship, we should store the Profile ID in the User table instead.
```go
// Profile belongs to User.
type Profile struct {
	ID     int64 `bun:",pk"`
	Lang   string
}

type User struct {
	ID      int64 `bun:",pk"`
	Name    string
        ProfileID string
	Profile *Profile `bun:"rel:has-one,join:profile_id=id"`
}
```
In this example, we need to create a foreign key, like `FOREIGN KEY ("profile_id") REFERENCES "profiles" ("id")`.
So maybe we can’t simply remove the logic that creates the foreign key for the has-one relationship. 
@dordieux 

If it's a belong-to relationship, we should store the user's id in the profile table.
```go
// Profile belongs to User.
type Profile struct {
	ID     int64 `bun:",pk"`
	Lang   string
        UserID string 
        User *User `bun:"rel:belong-to,join:user_id=id"`
}

type User struct {
	ID      int64 `bun:",pk"`
	Name    string
}
```
In this example, we need to create a foreign key, like `FOREIGN KEY ("user_id") REFERENCES "user" ("id")`


[...]

PR #1126, opened 2025-02-13
chore: fix has-one relation migrate

dordieux:
This is a fix for the PR: https://github.com/uptrace/bun/issues/1115.

It fixes a bug where the foreign key was incorrectly created on the table that has a 'has-one' relation.
I also modified the tests slightly to improve clarity.

### DIFF
--- a/internal/dbtest/inspect_test.go
+++ b/internal/dbtest/inspect_test.go
@@ -34,20 +34,23 @@ type Article struct {
 
 type Office struct {
 	bun.BaseModel `bun:"table:admin.offices"`
-	Name          string `bun:"office_name,pk"`
-	TennantID     string `bun:"publisher_id"`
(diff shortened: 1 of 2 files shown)


bevzzz:
From my understanding, `has-one` and `belongs-to` are 2 sides of 1 coin, that is, they are almost synonymous. The choice between them depends wholly on how the user chooses to model their data. 

Some may say (following the example from he docs) that "a user **has one** profile" and some that "this profiles **belongs to** a user"; they will then model their database accordingly. Provided they called `.WithForeignKeys()`, a foreign key should be created in either case.

Which is why I am not in favor of dropping this functionality for one of these relation types.

dordieux:
@bevzzz 
Thank you for your review.
As discussed in the original issue, I agree as well.
I'll close this for now and reconsider.
I appreciate your support!

PR #1129, opened 2025-02-14
chore: standardize behind error setting method

josephferrero:
I noticed that there were some methods setting the error on the query directly rather than using the provided setErr() method, so I standardized them all to use the setErr() method.


Aoang (approved):
LGTM

PR #1130, opened 2025-02-15
fix(create_table): avoid creating unintended foreign keys

### LINKED ISSUE #1115: `rel:has-one` on a relationship should not have foreign key creation

bevzzz:
Foreign keys should be created for has-one and belongs-to relations iff:
- None of the referencing columns is a primary keys
- The table is an m2m 'junction' table an all referencing columns are primary keys

See more details [in this comment](https://github.com/uptrace/bun/issues/1115#issuecomment-2661143967).

Closes #1115

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -1849,7 +1849,7 @@ func testNoForeignKeyForPrimaryKey(t *testing.T, db *bun.DB) {
 				bun.BaseModel `bun:"table:profiles"`
 				ID            string `bun:",pk"`
 				UserID        string
-			} `bun:"rel:belongs-to,join:id=user_id"`
+			} `bun:"rel:has-one,join:id=user_id"`
 		})(nil), dontWant: sqlschema.ForeignKey{
 			From: sqlschema.NewColumnReference("users", "id"),
 			To:   sqlschema.NewColumnReference("profiles", "user_id"),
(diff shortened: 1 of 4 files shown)


j2gg0s:
> Btw, do we squash PR commits or not?

I personally prefer squash commits, so that the master branch focuses only on the final result rather than the intermediate process.

However, this is not currently mandatory.

bevzzz:
> LGTM

I'm going to take that as a green light and merge this, @j2gg0s

PR #1132, opened 2025-02-17
fix: process embedded's struct field for table

### LINKED ISSUE #1125: Regression in v1.2.9 - Cannot scan relations when using struct embedding
### Description:
After upgrading from Bun v1.1.7 to v1.2.9, scanning relations into embedded structs fails with column recognition errors. A working implementation in v1.1.7 breaks in v1.2.9 due to missing column detection in embedded structs.

### Models:
```go
type User struct {
    Id string `bun:",pk"`

    bun.BaseModel `bun:"table:users"`
}

type Base struct {
    Id        string `bun:",pk"`
    CreatorId string

    bun.BaseModel `bun:"table:bases"`
    CreatedBy *User `bun:"rel:belongs-to,join:creator_id=id"`
}

type Enhanced struct {
    Base
    Extra string `bun:"-"`
}
```

### Query:
```go
err := db.NewSelect().
    Model((*Base)(nil)).
    Relation("CreatedBy").
    Scan(ctx, &Enhanced{})
```
### Error:
```
sql: Scan error on column index 2, name "created_by__id": 
bun: Enhanced does not have column "created_by__id"
```

### Expected Behavior:
Should successfully scan relation columns into embedded struct fields as in v1.1.7

### Actual Behavior:
Fails to recognize relation columns in embedded structs

### Additional Context:
 - Works correctly in v1.1.7
 - Breaks in v1.2.9
 - Error occurs when scanning into struct that embeds the base model with relations
 - Full reproduction available if needed

### Possible Related Changes:
This might be related to changes in struct scanning or relation handling for embedded structs between these versions.

### Why This Matters:
This regression breaks existing code that uses struct composition patterns, which is a common Go idiom. The error suggests the scanner isn't properly traversing embedded structs to find relation columns.

j2gg0s:
Close #1125


schema/table_test.go
```diff
@@ -92,6 +92,7 @@ func TestTable(t *testing.T) {
 		type Perms struct {
 			View   bool
 			Create bool
+			Model  *Model
```
vmihailenco, line 95: This is just a regular (JSON_ field, e.g. `Foo sql.NullBool` is just the same. Why do we need to treat it differently?

vmihailenco: Also, https://github.com/uptrace/bun/issues/1125 is about embedded structs, but this is not. `Perms` is not even a model...

j2gg0s: Fixed is #1137

PR #1135, opened 2025-02-19
fix(schema): determine whether a field is ambiguous with prefix

j2gg0s:


j2gg0s:
I didn’t receive specific feedback, but based on the context, we should include the prefix when making the determination?

@vmihailenco

vmihailenco:
@j2gg0s you're probably right, but can we add a test to the table_test.go ?

Issue #1107, opened 2025-01-13
Regression - panic scanning model with custom PK type and has-many field


NathanBaulch:
I'm no longer able to scan a model with a `has-many` relation included when using PK types that implement `driver.Valuer`. I'm using https://github.com/oklog/ulid for all PKs in my model.
This regression was introduced by #1080 in `v1.2.7` and seen previously in #779. cc @thecampagnards @Tiscs 

```
panic({0x207b680?, 0xc0000672c0?})
        .../scoop/apps/go/current/src/runtime/panic.go:785 +0x132
github.com/uptrace/bun.baseValues.func1({0x22fefe0?, 0xc0003d2280?, 0x132ffbe0108?})
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/model_table_has_many.go:140 +0x136
github.com/uptrace/bun.visitField({0x22fefe0?, 0xc0003d2280?, 0xc000500008?}, {0xc0004a7840, 0x0, 0x1}, 0xc0008930b8)
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/util.go:38 +0xcf
github.com/uptrace/bun.walk({0x22fefe0?, 0xc0003d2280?, 0x0?}, {0xc0004a7840, 0x0, 0x1}, 0xc0008930b8)
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/util.go:25 +0x9f
github.com/uptrace/bun.baseValues({0x26def00, 0xc000896000}, {0xc00009a740, 0x1, 0x1})
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/model_table_has_many.go:137 +0x1c7
github.com/uptrace/bun.newHasManyModel(0xc0007b1c70)
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/model_table_has_many.go:28 +0x75
github.com/uptrace/bun.(*relationJoin).manyQuery(0xc0007b1c70, 0xc0007b8960)
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/relation_join.go:57 +0x25
github.com/uptrace/bun.(*relationJoin).selectMany(0x92983a?, {0x26d0850, 0xc00055f6e0}, 0x0?)
        .../go/pkg/mod/github.com/uptrace/bun@v1.2.7/relation_join.go:49 +0x25
[...]

PR #1139, opened 2025-02-25
Fix handle ptr pk in indirectAsKey

Aoang:


model_table_has_many.go
```diff
@@ -172,5 +172,18 @@ func indirectAsKey(field reflect.Value) interface{} {
...
+	if valuer, ok := field.Interface().(driver.Valuer); ok {
+		if v, err := valuer.Value(); err == nil {
+			switch reflect.TypeOf(v).Kind() {
+			case reflect.Array, reflect.Chan, reflect.Func,
+				reflect.Map, reflect.Ptr, reflect.Slice, reflect.UnsafePointer:
+				break
+			default:
+				return v
+			}
+		}
```
vmihailenco, line 185: I think we should always use the value returned by `driver.Valuer` even if it is not comparable.

I suggest to do this:

```
val, err := valuer.Value()
if err != nil {
    slog.Error(...)
}
return val // might be nil
```

Aoang: Maybe we don't need to log because the panic will come later

vmihailenco: It might panic on nil value (or not), but you won't see the error message in app logs which will make debugging slightly more complicated... Up to you.

vmihailenco (approved):

model_table_has_many.go
```diff
@@ -172,5 +172,11 @@ func indirectAsKey(field reflect.Value) interface{} {
 	if field.IsNil() {
 		return nil
 	}
+
+	if valuer, ok := field.Interface().(driver.Valuer); ok {
+		v, _ := valuer.Value() // No need to record logs, will panic later
```
j2gg0s, line 177: Do we need to be compatible with cases similar to #1107? 
Suppose the user is using *ULID,  if we don’t handle this specifically, 
it will panic because [16]byte is unhashable in Go.

Aoang: We only deal with pointer-based `driver.Valuer` implementation. What do you think?

> https://go.dev/play/p/kY2YJUrEx8r

j2gg0s: In your test case, the struct itself is not comparable, but the result returned by Value is comparable.
So it would have panicked before, but now it won’t.

In the example I provided, the struct(array) itself is comparable, 
but the result returned(slice) by Value is not comparable.
So it wouldn’t have panicked before, but now it will.
```go
type ULID [16]byte

func (id *ULID) Scan(any) error               { return nil }
func (id *ULID) Value() (driver.Value, error) { return []byte{}, nil }

type User struct {
	ID    *ULID `bun:",pk"`
	Name  string
	Posts []*Post `bun:"rel:has-many,join:id=user_id"`
}

type Post struct {
	PostID string `bun:",pk"`
	UserID *ULID
}
```

Maybe I misunderstood, but this isn’t related to whether it’s pointer-based, is it?

Aoang: Interfaces like `driver.Valuer`, `json.Marshaler`, and most other encoding interfaces are generally implemented directly on values rather than pointers.

Protobuf generated code has another characteristic, [`DoNotCopy`](https://github.com/protocolbuffers/protobuf-go/blob/master/internal/pragma/pragma.go#L29) , in addition to [`DoNotCompare`](https://github.com/protocolbuffers/protobuf-go/blob/master/internal/pragma/pragma.go#L22) , this imposes a strict requirement that all methods must be pointer-based when extending the generated code.

Deviating from this best practice will inevitably lead to more issues. For example, just like now.

I will fix it later.

Issue #1134, opened 2025-02-18
Data race in v1.2.10


wavded:
In v1.2.9 we noticed panics when running tests that used `ScanAndCount`, we no longer experience that in v1.2.10 which is great, but we noticed this race condition exists in v1.2.10. This race condition does not exist in v1.2.8.

```
go test -race
==================
WARNING: DATA RACE
Write at 0x00c00020b558 by goroutine 47:
  github.com/uptrace/bun.(*relationJoin).applyTo()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/relation_join.go:41 +0x237
  github.com/uptrace/bun.(*SelectQuery).appendQuery.func1()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:540 +0x34
  github.com/uptrace/bun.(*SelectQuery)._forEachInlineRelJoin()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:461 +0xc2
  github.com/uptrace/bun.(*SelectQuery).forEachInlineRelJoin()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:453 +0x7e
  github.com/uptrace/bun.(*SelectQuery).appendQuery()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:539 +0x3f7
  github.com/uptrace/bun.(*SelectQuery).AppendQuery()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:513 +0xd8
  github.com/uptrace/bun.(*SelectQuery).scanResult()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:884 +0x49c
  github.com/uptrace/bun.(*SelectQuery).Scan()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:849 +0x124
  github.com/uptrace/bun.(*SelectQuery).scanAndCountConcurrently.func1()
      /var/lib/jenkins/go/pkg/mod/github.com/uptrace/bun@v1.2.10/query_select.go:988 +0x125

Previous write at 0x00c00020b558 by goroutine 48:
  github.com/uptrace/bun.(*relationJoin).applyTo()
[...]

PR #1146, opened 2025-03-14
fix: relation join data race

Aoang:
Fix a potential data race condition that could occur during relation joins. The issue stemmed
from concurrent access to the `tableModel` during the join operation. To resolve this, the `clone()` method has been implemented for the `TableModel` interface and its concrete implementations (`hasManyModel`, `m2mModel`, `sliceTableModel`, `structTableModel`). This ensures that each concurrent operation works with its own copy of the `tableModel`, preventing race conditions.

Additionally, the Makefile has been updated to include race condition detection during testing by adding the `RACETEST=1` environment variable to the `go test` command.

Fix: https://github.com/uptrace/bun/issues/1134

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -2006,6 +2006,8 @@ func testWithPointerPrimaryKeyHasManyWithDriverValuer(t *testing.T, db *bun.DB) 
 	require.Len(t, program.Items, 2)
 }
 
+// TODO: Investigate and fix race conditions in other dialects for this test.
+// See https://github.com/uptrace/bun/pull/1146
 func testRelationJoinDataRace(t *testing.T, db *bun.DB) {
 	if db.Dialect().Name().String() != pgName {
 		return
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -6,6 +6,7 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -35,7 +36,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var ctx = context.TODO()
+var (
+	ctx        = context.TODO()
+	isRaceTest = os.Getenv("RACETEST") != ""
+)
 
 const (
 	pgName        = "pg"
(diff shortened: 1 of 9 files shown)


internal/dbtest/inspect_test.go
```diff
@@ -73,6 +73,10 @@ type Journalist struct {
 }
 
 func TestDatabaseInspector_Inspect(t *testing.T) {
+	if isRaceTest {
+		t.Skip("TODO: fix race test")
+	}
```
bevzzz, line 78: Does the code in the test cause a race condition?
I seem to be a bit out of context here -- could you please add a short description of where and why the race happens to the PR?

internal/dbtest/db_test.go
```diff
@@ -2000,3 +2005,40 @@ func testWithPointerPrimaryKeyHasManyWithDriverValuer(t *testing.T, db *bun.DB)
 	require.NoError(t, err)
 	require.Len(t, program.Items, 2)
 }
+
+func testRelationJoinDataRace(t *testing.T, db *bun.DB) {
+	if db.Dialect().Name().String() != pgName {
+		return
```
bevzzz, line 2011: Is the test PG-specific, or do we skip other dialects because we only need to test with one?

If it's the latter, would it be clearer to just make this a separate test altogether?
Like [`TestMssqlMerge`](https://github.com/uptrace/bun/blob/7fc6639dea51f98c1f87c24a82609844edc74252/internal/dbtest/mssql_test.go#L9), which explicitly uses only 1 dialect.

internal/dbtest/inspect_test.go
Aoang: This PR addresses the issue described in https://github.com/uptrace/bun/issues/1134 .

To properly test the fix, we ideally want to ensure that our tests can detect race conditions. The simplest way to achieve this is to run go test -race. Alternatively, without the race detector, we would have to manually introduce multiple goroutines in our tests to reliably trigger these conditions, which can be complex and less comprehensive.

However, some existing tests in the codebase currently fail when the race detector is enabled. These failures are not directly related to the fix for #1134 and need to be investigated and addressed separately. Running go test -race on the entire suite would block progress on this PR.

Therefore, the t.Skip statement temporarily skips these failing test cases only when running with the -race flag. This allows us to focus on verifying the fix for the primary issue without being immediately blocked by unrelated race conditions elsewhere in the test suite. A follow-up issue/PR will be created to address the skipped race tests.

internal/dbtest/db_test.go
Aoang: Similar to the previous issue, this test isn't theoretically specific to PostgreSQL. However, these tests currently exhibit unrelated failures under the race detector, making it difficult to isolate the intended behavior. Perhaps I should add a TODO note here to indicate that these tests need to be revisited and fixed for other dialects in the context of race condition detection?

bevzzz: Agree, a comment next to that if-block explaining why we only test for a single dialect now and how we should address that in the future would be good!

j2gg0s:
LGTM, thanks for your contribution!

Issue #1144, opened 2025-03-09
Adding interfaces for query methods


DataHearth:
Hey!

We're using `bun` in our projects, and I was wondering if it could be possible to add interfaces to queries like `NewSelect`, `NewUpdate`, etc...
If this ok, I can submit a PR adding those interfaces 😃 .

j2gg0s:
Can you introduce your specific use cases?

I tend to wait until there is a concrete need before making this change.

DataHearth:
Sure !

We're doing extensive testing on our codebase. Requiring a database to run our tests significantly slows down the execution and requires additional tooling for injecting/cleaning up data.

With interfaces, we could operate mocks using the `mockery` + `testutils` suite and leverage their features to add additional tests without requiring additional code and adding complexity.

We still want to rely on the database for E2E tests though.

j2gg0s:
Can we use sql/driver to implement mock?

DataHearth:
On your side, this is not needed. This would be DX pull request destined to improve the overall usability of bun inside tests (this shouldn't impact current logical behavior or whatsoever).

On our side though, this would be difficult as we don't expose nor pass down the `database/sql.DB` struct. We only propagate your interface `bun.IDB`. This is not enough as deeper struct like `bun.NewSelect` doesn't rely on this interface and returns a `QuerySelect` struct on each downstream method.  

My proposition is, instead of returning a struct `QueryXXX`, returning an interface dedicated to each `QueryXXX` reflecting its methods.

j2gg0s:
Okay, I’m just concerned about the workload and the benefits.

DataHearth:
For the benefits, it's adding a smoother DX overall IMO. For the workload, I guess the first PR I'll do will be slightly big. For the maintainability, whenever a method is added/removed/changed, "just" updating the corresponding interface(s). I'll let you decide if this is too much for you (team) or not 😄. I can of course contribute to this after the first implementation is done on my free time ^^.

j2gg0s:
@DataHearth 
Thanks for your contribution.
I’m just afraid this will take up your time.
For Bun, I think this is obviously beneficial.

You don’t need to wait for all the work to be done before submitting a PR.
We dont mind WIPS.

DataHearth:
@j2gg0s Hey ! I've started working on it. It way more complicated than expected. I'm gonna do a bunch or PRs at the time to avoid a big chunky PR impossible to review.
[...]

PR #1148, opened 2025-03-20
feat(queries): add interfaces - part 1

DataHearth:
Part of #1144

Every query structures inside `bun` uses the structure itself to function properly. While this behavior is good, it is not flexing for testing and/or re-implementing certain features.

Adding interfaces add this flexibility.

This PR is part of a group to implement interfaces where necessary to increase the DX.

### DIFF
--- a/db.go
+++ b/db.go
@@ -126,7 +126,7 @@ func (db *DB) NewValues(model any) IValuesQuery {
 	return NewValuesQuery(db, model)
 }
 
-func (db *DB) NewMerge() *MergeQuery {
+func (db *DB) NewMerge() IMergeQuery {
 	return NewMergeQuery(db)
 }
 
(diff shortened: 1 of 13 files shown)


j2gg0s:
@DataHearth 
Thank you very much for your work.

I’m sorry, but I may not be able to participate until next week or the week after. My work has been a bit busy recently.

vmihailenco:
This looks good, but I have 2 questions to discuss:

1. This is a backwards incompatible change. Do we plan to release this as v2.0? What other changes do we plan to include? I'd like to change the behavior of `bun.In` for v2.0.

2. What other changes are planned? I'd like to better understand how many APIs we're going to break before this is done.

DataHearth:
@vmihailenco Hey ! For this kind of change, only the query methods I guess. I'll see when I'll do the rest of them. 

It is not compatible only if you're using struct's fields I think. We're using bun internally, this would not affect our workflow for example. But I do agree that if you're using bun for more than just querying your database, this would break a few things.

@j2gg0s Sure ! Take your time 😄. I did this first PR to give you an example of the migration to interfaces. When discussions are well advanced and you validate this kind of changes, we could move forward if wanted 👍.

PR #828, opened 2023-05-09
Add option for unicode paramteter in appendstring function

MohammadAnsari98:
This pull request is for https://github.com/uptrace/bun/issues/824 issue

Issue #824, opened 2023-05-04
MSSQL performance is degraded by this pull request


babashark:
https://github.com/uptrace/bun/pull/768

We do not need 'N' if you are using collation Latin1_General_100_CI_AI_SC_UTF8.

Adding the 'N' make sql server failed to use index seek, caused a lot of dead lock and performance issues.

I think we should add option to use 'N' or not.

<img width="1286" alt="image" src="https://user-images.githubusercontent.com/127459243/236183286-48399bda-73a2-4f02-bbb0-2cbefd616158.png">

vmihailenco:
It should be fairly straightforward to add this as an option to dialect: https://github.com/uptrace/bun/blob/master/dialect/mssqldialect/dialect.go#L130-L134

vmihailenco:
This should be a dialect-wide option, e.g.

```go
dialect := mssqldialect.New(mssqldialect.WithUnicode(false))
```

PR #1157, opened 2025-03-26
Add option for unicode to mssql dialect

bjongbloedt:
Add an option to support non unicode collated mssqlserver dialect.

Fixes [#824](https://github.com/uptrace/bun/issues/824), and based on [PR-828](https://github.com/uptrace/bun/pull/828), but with a global option instead of a local option


j2gg0s:
Thank you for your work.

Issue #1154, opened 2025-03-25
CamelCase to snake_case is broken for acronyms (ID, HTML, etc.)


jeffreydwalter:
`UserID` is getting converted to `user_i_d`

j2gg0s:
Can we refer to the existing logic of golangci-lint and similar tools?

jeffreydwalter:
I don't understand your comment.

j2gg0s:
Sorry about that, I was debugging another program last night and it made my head a bit dizzy.

The implementation of this requirement might be similar to [Initialisms](https://go.dev/wiki/CodeReviewComments#initialisms).
This kind of feature is commonly seen in linters, for example [revive](https://github.com/mgechev/revive/blob/master/lint/name.go#L9).
I think we can borrow some of the logic from them.

We also need to consider compatibility with past behavior, maybe at least keep it as a configurable option.

j2gg0s:
I think the specific implementation approach is acceptable as long as the logic aligns with the habits of most people.

However, this is indeed a difficult task, especially considering that there might be users of the database who use other languages. The conventions around initialisms can vary greatly between languages.

This is not a mandatory requirement, because we allow users to define names through tags.

If you want to change the logic here, please maintain compatibility and provide an option through a configuration setting to allow users to enable the new feature. I really don’t want users to find that code that used to work is no longer working after an upgrade.

bevzzz:
> If you're willing to break backwards compatibility

I agree with @j2gg0s -- for a feature like this b/c should definitely be maintained. After all, `user_i_d` is not a bug that would _break an application_, it's just an odd-looking column name.

Rather than fixing `internal.Underscore` we should declare some `Canonicalizer` interface and user that to convert struct/field names to table/column names. Similarly to how `TableNameInflector` is defined:

https://github.com/uptrace/bun/blob/98cbfbfe0ac64b0c4b309848052b3c516160b335/schema/table.go#L32-L34

While `internal.Underscore` should stay the default option for b/c, we will be free to provide another canonicalizer that will respect abbreviations. Something like what `revive` has, as mentioned above.

Opting-in should be really simple.

```go
schema.SetCanonicalizer(schema.UnderscoreAbbr)
```

We could even go a step further and allow specifying initialisms that canonicalizer should recognize:

```go
schema.SetCanonicalizer(schema.UnderscoreAbbrCustom("ABBA", "BYOB"))
```

jeffreydwalter:
[...]

PR #1163, opened 2025-04-01
chore: add test case for acronyms

bevzzz:
Related #1154 

User's claims that CamelCase conversion is broken and that

> UserID is getting converted to user_i_d

This PR adds a test case to TestQuery suit to verify that, when running a CreateNewTable query, table/column names containing acronyms are converted to snake_case without breaking the acronyms apart, i.e. UserID -> user_id.


bevzzz:
CI is failing because I only generate the test snapshot for `pg`, so the other dialects don't have one to compare their result to.

Will fix now.

Aoang (approved):
LGTM

PR #1165, opened 2025-04-03
chore: rename migrate.comment to migrate.Unimplemented

bevzzz:
Another follow up to [this comment](https://github.com/uptrace/bun/pull/926#issuecomment-2758161051) in #926:

> Also, it would be handy for migrate.comment to be an exported type. As it is, I had to use reflection to detect them.

All other operations are exported types with exported fields. We can export comment as well to make 'migrate' package more extendable / interoperable.


j2gg0s:
LGTM

PR #1168, opened 2025-04-14
Update query_raw.go

yestoolbox:


Aoang:
Thanks again for all the recent pull requests you've submitted!

We've noticed that several of them focus on the same type of change: adding explicit type parameters to calls of the generic function `setCommentFromContext`, like changing `setCommentFromContext(ctx, q)` to `setCommentFromContext[*SomeType](ctx, q)`.

We understand the likely motivation behind this change is to improve type inference support for certain IDEs that might not automatically deduce the correct type parameter in this context.

To help streamline the review process and maintain a cleaner commit history, would you be able to consolidate all these related changes into a single pull request?

In the description of that consolidated PR, could you please clearly explain the reasoning behind the change? Mentioning the specific IDE challenges this aims to address would be very helpful for context during the review.

This will make it much easier for us to review and discuss the proposed changes effectively as a whole.

Thanks again for your contributions and for your help with this! We look forward to seeing the consolidated PR.

PR #1188, opened 2025-04-23
feat(migrations): support Go templates in SQL migrations

rionthys:
## Summary

This PR adds support for rendering `.sql` migration files as Go templates via `text/template`.  
You can now inject dynamic data into your SQL using a new migrator option: `WithTemplateData`.

## Usage

Pass your data using the new option:

```go
templateData := map[string]string{
    "Prefix": "example_",
}

migrator := migrate.NewMigrator(db, migrations.Migrations, migrate.WithTemplateData(templateData))
```

And in your .sql file:
```sql
CREATE TABLE {{.Prefix}}users (
    id BIGINT PRIMARY KEY
);

--bun:split

ALTER TABLE {{.Prefix}}users ADD COLUMN name VARCHAR(100);
```

## Changes
- `WithTemplateData(data any)` provides template data to be rendered inside .sql migration files
- `internalMigrationFunc` introduced to wrap `MigrationFunc` with template context
- `.sql` files are parsed using text/template before execution
- Template rendering is fully optional and backward-compatible
- Updated example: `example/migrate/main.go`
- Added test template in: `example/migrate/migrations/*.sql`

### DIFF
--- a/migrate/migration.go
+++ b/migrate/migration.go
@@ -58,21 +58,13 @@ func wrapMigrationFunc(fn MigrationFunc) internalMigrationFunc {
 	}
 }
 
-func prepareTemplate(f io.Reader, templateData any) (*bytes.Buffer, error) {
-	contents, err := io.ReadAll(f)
-	if err != nil {
-		return nil, err
-	}
-
+func renderTemplate(contents []byte, templateData any) (*bytes.Buffer, error) {
 	tmpl, err := template.New("migration").Parse(string(contents))
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse template: %w", err)
 	}
 
 	var rendered bytes.Buffer
-	if templateData == nil {
-		templateData = struct{}{}
-	}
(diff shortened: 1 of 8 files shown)


migrate/migration.go
```diff
@@ -35,23 +36,57 @@ func (m Migration) IsApplied() bool {
...
+	}
+
+	tmpl, err := template.New("migration").Parse(string(contents))
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse template: %w", err)
+	}
+
+	var rendered bytes.Buffer
+	if templateData == nil {
+		templateData = struct{}{}
```
vmihailenco, line 74: There is no point to execute template if the templateData` is nil. Instead let's check for nil and bypass `prepareTemplate` competely.

rionthys: Implemented, thanks for the note!

vmihailenco (approved):

Aoang (approved):
LGTM

Issue #1166, opened 2025-04-10
[Tracking issue]: AutoMigrator Improvements


bevzzz:
- [x] Replace `ordered.Map` with slices in public APIs (#1164)
- [x] Make `migrate.comment` and exported type (#1165)
- [x] Exclude foreign keys from database inspection (#1162)
- [ ] Exclude tables/foreign keys that match a pattern ([comment](https://github.com/uptrace/bun/pull/926#issuecomment-2791500903))
- [ ] Avoid redundant migrations for `autoincrement` column (#1159)
- [ ] Add option to generate migration file names ([comment](https://github.com/uptrace/bun/issues/1159#issuecomment-2791606202))

bevzzz:
> And I needed to be able to set the directory and the output filenames.
@bgrant0607 wrt the **output filenames**: currently the autogenerated one'd look like so `20250508184622_public_auto.sql`. If there was an option to control filenames would you expect it to:
- let you create the whole filename?
- let you create just the "description" part and always append the timestamp?
- some combination of both? e.g. `WithFileNameFunc(fn func() string, timestamp boolean))`

Are there any parameters that would be useful to pass to that `fn`?

I'd appreciate your input on that.

cc @j2gg0s @Aoang

j2gg0s:
> > And I needed to be able to set the directory and the output filenames.
> 
> [@bgrant0607](https://github.com/bgrant0607) wrt the **output filenames**: currently the autogenerated one'd look like so `20250508184622_public_auto.sql`. If there was an option to control filenames would you expect it to:
> 
> * let you create the whole filename?
> * let you create just the "description" part and always append the timestamp?
> * some combination of both? e.g. `WithFileNameFunc(fn func() string, timestamp boolean))`
> 
> Are there any parameters that would be useful to pass to that `fn`?
> 
> I'd appreciate your input on that.
> 
> cc [@j2gg0s](https://github.com/j2gg0s) [@Aoang](https://github.com/Aoang)

- Add an options parameter that allows users to pass in a function which takes input and returns the filename.
  - The current input is schema; whether other parameters should be included needs to be carefully considered and decided, in order to avoid future compatibility issues.
- Define the current behavior and other reasonable defaults as functions, and use one of them as the default option.
- As long as there’s no risk involved, I lean toward offering maximum flexibility.

bevzzz:
@snoweuph thank you for reporting this -- that's an actual bug you found there. I will add a fix.

giautm:
[...]

PR #1190, opened 2025-04-29
feat(automigrate): exclude tables using LIKE pattern

bevzzz:
Related to #1166 

This PR extends AutoMigrator / Inspector `ExcludeTables` option to allow patters that use LIKE-operator syntax.

```go
auto := migrate.NewAutoMigrator(db, migrate.WithExcludeTableLike("%-exclude-%"))
```

### Not in scope

Foreign keys are excluded based on their definition, not name, so WithExcludeForeignKeys option does not receive the -Like variant.

### DIFF
--- a/internal/dbtest/docker-compose.yaml
+++ b/internal/dbtest/docker-compose.yaml
@@ -1,32 +1,32 @@
 services:
-  mysql8:
-    image: mysql:8.0
-    environment:
-      - MYSQL_DATABASE=test
-      - MYSQL_USER=user
-      - MYSQL_PASSWORD=pass
-      - MYSQL_ROOT_PASSWORD=pass
-    ports:
-      - 3306:3306
-    healthcheck:
-      test: ['CMD', 'mysqladmin', 'ping']
-      timeout: 5s
-      interval: 10s
-      retries: 3
-  mysql5:
-    image: mysql:5.7
-    environment:
-      - MYSQL_DATABASE=test
-      - MYSQL_USER=user
-      - MYSQL_PASSWORD=pass
-      - MYSQL_ROOT_PASSWORD=pass
-    ports:
-      - 53306:3306
-    healthcheck:
-      test: ['CMD', 'mysqladmin', 'ping']
-      timeout: 5s
-      interval: 10s
-      retries: 3
+  # mysql8:
+  #   image: mysql:8.0
+  #   environment:
+  #     - MYSQL_DATABASE=test
+  #     - MYSQL_USER=user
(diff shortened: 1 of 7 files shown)


migrate/sqlschema/inspector.go
```diff
@@ -57,12 +68,23 @@ func WithSchemaName(schemaName string) InspectorOption {
...
-// from the reported schema state.
+// WithExcludeTables forces inspector to exclude tables from the reported schema state.
 // It works in append-only mode, i.e. tables cannot be re-included.
+//
+// All tables are additionally appended to WithExcludeTablesLike
+// for forward-compatibility, allowing dialects to ignore deprecated ExcludeTables.
 func WithExcludeTables(tables ...string) InspectorOption {
 	return func(cfg *InspectorConfig) {
 		cfg.ExcludeTables = append(cfg.ExcludeTables, tables...)
+		cfg.ExcludeTablesLike = append(cfg.ExcludeTablesLike, tables...)
```
bevzzz, line 79: Because patterns without wildcards are valid, we also deprecate `sqlschema.InspectorConfig.ExcludeTables` and encourage dialects (pgdialect for now) to use ExcludeTablesLike only.
 
To make this possible `sqlschema.WithExcludeTables` appends to both ExcludeTables and ExcludeTablesLike list.

j2gg0s: Can we directly reuse this API while allowing users to pass both actual table names and regex-based table names?
That way, we only need to modify the underlying implementation and can continue using WithExcludeTables.

internal/dbtest/docker-compose.yaml
```diff
@@ -1,32 +1,32 @@
...
-      - MYSQL_PASSWORD=pass
-      - MYSQL_ROOT_PASSWORD=pass
-    ports:
-      - 53306:3306
-    healthcheck:
-      test: ['CMD', 'mysqladmin', 'ping']
-      timeout: 5s
-      interval: 10s
-      retries: 3
+  # mysql8:
```
j2gg0s, line 2: What is the purpose of this change? I don’t quite understand it for now.

bevzzz: My bad! I commented those out for slightly faster test runs locally and must've committed those accidentally.
Thanks for catching it.

migrate/sqlschema/inspector.go
bevzzz: Yes, that's a great suggestion!

bevzzz:
@j2gg0s addressed your feedback, thanks for the review!

PR #1192, opened 2025-04-29
fix(automigrate): improve migrations on `autoincrement` columns

### LINKED ISSUE #1159: AutoMigrate doesn't fully handle autoincrement for postgres
This is to make you aware of the issue. I don't have a full fix yet.

autoincrement fields map to serial (for int) and bigserial (for int64) pseudo-types.

serial and bigserial are kind of like macros that expand into integer and bigint.

https://www.postgresql.org/docs/9.1/datatype-numeric.html

"The data types serial and bigserial are not true types, but merely a notational convenience
for creating unique identifier columns (similar to the AUTO_INCREMENT property supported by some other databases)."

int64 autoincrement is generated as bigserial, which in postgres automatically translates to:
```
CREATE SEQUENCE tablename_colname_seq;
CREATE TABLE tablename (
	colname bigint DEFAULT nextval('tablename_colname_seq')::regclass NOT NULL
);
ALTER SEQUENCE tablename_colname_seq OWNED BY tablename.colname;
```

This interferes with the diff code in multiple ways:
1. Changes the type
2. Adds NOT NULL
3. Adds a DEFAULT

I'm not sure whether the diff code pays attention to the ALTER SEQUENCE yet.

https://github.com/uptrace/bun/blob/master/migrate/diff.go#L277
```
func (d detector) equalColumns(col1, col2 sqlschema.Column) bool {
	return d.cmpType(col1, col2) &&
		col1.GetDefaultValue() == col2.GetDefaultValue() &&
		col1.GetIsNullable() == col2.GetIsNullable() &&
		col1.GetIsAutoIncrement() == col2.GetIsAutoIncrement() &&
		col1.GetIsIdentity() == col2.GetIsIdentity()
}
```

The comparison function passed for postgres doesn't tolerate the type change.

bevzzz:
Closes #1159 
Related #1166 

As per the related issue, AutoMigrator creates superficial migrations for `autoincrement` columns, attempting to change type to "SERIAL" (for Postgres) and drop NOT NULL.

Because SERIAL is a create-table alias for INTEGER + generated sequence, Postgres dialect should report the two types as equivalent. Exciting addition 🎉 : when changing a column type from INTEGER to SERIAL, the dialect will also add SQL to **create a backing sequence** first and **generate default value** from it. Previously we were not handling this case correctly.

Furthermore, fields with `autoincrement` tag should have `f.NotNull = true`, because an autoincremented field cannot contain null values.

### DIFF
--- a/dialect/pgdialect/alter_table.go
+++ b/dialect/pgdialect/alter_table.go
@@ -1,6 +1,7 @@
 package pgdialect
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
@@ -57,6 +58,11 @@ func (m *migrator) AppendSQL(b []byte, operation interface{}) (_ []byte, err err
 	case *migrate.DropUniqueConstraintOp:
 		b, err = m.dropConstraint(fmter, appendAlterTable(b, change.TableName), change.Unique.Name)
 	case *migrate.ChangeColumnTypeOp:
+		// If column changes to SERIAL, create sequence first.
+		// https://gist.github.com/oleglomako/185df689706c5499612a0d54d3ffe856
+		if !change.From.GetIsAutoIncrement() && change.To.GetIsAutoIncrement() {
+			change.To, b, err = m.createDefaultSequence(fmter, b, change)
+		}
 		b, err = m.changeColumnType(fmter, appendAlterTable(b, change.TableName), change)
 	case *migrate.AddForeignKeyOp:
 		b, err = m.addForeignKey(fmter, appendAlterTable(b, change.TableName()), change)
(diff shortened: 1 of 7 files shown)


bevzzz:
Generating sequence for an integer column:

```sql
SELECT MAX('incremented') FROM "change_me_own_type"
CREATE SEQUENCE change_me_own_type_incremented_seq START WITH 1 OWNED BY change_me_own_type.incremented;
ALTER TABLE "public"."change_me_own_type" ALTER COLUMN "incremented" SET NOT NULL, ALTER COLUMN "incremented" SET DEFAULT nextval('change_me_own_type_incremented_seq'::regclass)
```

j2gg0s:
LGTM

Issue #1205, opened 2025-06-05
CREATE TABLE query is non-deterministic due to foreign key order


shota3506:
Calling CreateTableQuery.AppendQuery can produce a non-deterministic SQL query when the table includes foreign keys.

The root cause is that appendFKConstraintsRel iterates over a map to append foreign key constraints. Since the iteration order of maps is not guaranteed in Go.
I hope it returns the same query every time because I take a query snapshot in my project.

PR #1206, opened 2025-06-05
fix: sort foreign keys by name to ensure deterministic query

shota3506:
This PR fixes https://github.com/uptrace/bun/issues/1205

This change introduces sorting for the foreign key constraints before they are appended to the query string.

The performance impact of sorting is expected to be negligible, as the number of foreign keys on a single table is typically small.
This approach is consistent with other parts of the codebase, [CreateTableQuery.appendUniqueConstraints](https://github.com/uptrace/bun/blob/a76da0dfa562bd6e3eb1b330de954c86f8e4daf7/query_table_create.go#L273-L279), which already sorts keys to ensure deterministic output.

### DIFF
--- a/query_table_create.go
+++ b/query_table_create.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


Aoang (approved):
LGTM

PR #1207, opened 2025-06-06
refactor: use slices package for sorting

shota3506:
This PR migrates sorting logic from the legacy sort package to the modern slices package.

[Sorting for slices of struct values []migrate.Migration](https://github.com/uptrace/bun/blob/a76da0dfa562bd6e3eb1b330de954c86f8e4daf7/migrate/migration.go#L327-L337) has been intentionally excluded from this change, as slices.SortFunc copies the entire struct on each comparison.
Let me know if it's okay to ignore this concern.

### DIFF
--- a/model_map.go
+++ b/model_map.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"database/sql"
 	"reflect"
-	"sort"
+	"slices"
 
 	"github.com/uptrace/bun/schema"
 )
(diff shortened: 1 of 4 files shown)


j2gg0s:
Thanks for the contribution. Could you rebase it on the latest master branch when you have a moment?

shota3506:
@j2gg0s I've rebased

PR #1214, opened 2025-06-14
fix: restore q.limit check

vmihailenco:
This is follow-up for https://github.com/uptrace/bun/pull/1119

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -979,20 +979,25 @@ func (q *SelectQuery) scanAndCountConcurrently(
 	var mu sync.Mutex
 	var firstErr error
 
+	// FIXME: clone should not be needed, because the query is not modified here
+	// and should not be implicitly modified by the Bun lib.
 	countQuery := q.Clone()
 
-	wg.Add(1)
-	go func() {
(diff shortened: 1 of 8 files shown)


query_select.go
```diff
@@ -979,20 +979,25 @@ func (q *SelectQuery) scanAndCountConcurrently(
...
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-
-		if err := q.Scan(ctx, dest...); err != nil {
-			mu.Lock()
-			if firstErr == nil {
-				firstErr = err
+	// Don't scan results if the user explicitly set Limit(-1).
+	if q.limit >= 0 {
```
j2gg0s, line 987: In ScanAndCount: when offset == 0 && limit == 0, 
it performs only a scan query and returns both the results and the count.

Now there’s a request to add logic: if offset < 0, it should perform only the count operation.
Wouldn’t it be better to move this logic up into ScanAndCount?

Also, could you briefly explain why limit might be less than 0? I imagine not everyone is familiar with that.

PR #1221, opened 2025-06-19
fix: pgdialect range

thecampagnards:
Hello,
I suggest a fixes on `pgdialect.Range`, on a pg14.
Doc: https://www.postgresql.org/docs/current/rangetypes.html
Thx !

### DIFF
--- a/dialect/pgdialect/range.go
+++ b/dialect/pgdialect/range.go
@@ -1,138 +1,149 @@
 package pgdialect
 
 import (
+	"bytes"
 	"database/sql"
+	"database/sql/driver"
 	"fmt"
-	"io"
-	"time"
+	"reflect"
 
-	"github.com/uptrace/bun/internal"
 	"github.com/uptrace/bun/schema"
 )
 
 type MultiRange[T any] []Range[T]
 
-type Range[T any] struct {
-	Lower, Upper           T
-	LowerBound, UpperBound RangeBound
+var (
+	_ sql.Scanner   = (*MultiRange[any])(nil)
+	_ driver.Valuer = (*MultiRange[any])(nil)
+)
+
+func (m *MultiRange[T]) Scan(anySrc any) (err error) {
+	return Array(m).Scan(anySrc)
 }
 
-type RangeBound byte
+func (m MultiRange[T]) Value() (driver.Value, error) {
+	return m.String(), nil
+}
+
+func (m MultiRange[T]) String() string {
+	if len(m) == 0 {
+		return "{}"
+	}
+	var b []byte
+	b = append(b, '{')
+	for _, r := range m {
+		b = append(b, unquote(appendElem(nil, r))...)
+		b = append(b, ',')
+	}
+	b = append(b[:len(b)-1], '}')
+	return string(b)
+}
+
+type Range[T any] struct {
+	Lower, Upper T
+	LowerBound   RangeLowerBound
+	UpperBound   RangeUpperBound
+}
+
+var (
+	_ driver.Valuer = (*Range[any])(nil)
+	_ sql.Scanner   = (*Range[any])(nil)
+)
+
+type RangeLowerBound byte
+type RangeUpperBound byte
 
 const (
-	RangeBoundInclusiveLeft  RangeBound = '['
-	RangeBoundInclusiveRight RangeBound = ']'
-	RangeBoundExclusiveLeft  RangeBound = '('
-	RangeBoundExclusiveRight RangeBound = ')'
+	RangeBoundExclusiveLeft  RangeLowerBound = '('
+	RangeBoundExclusiveRight RangeUpperBound = ')'
+	RangeBoundInclusiveLeft  RangeLowerBound = '['
+	RangeBoundInclusiveRight RangeUpperBound = ']'
+
+	RangeBoundDefaultLeft  = RangeBoundInclusiveLeft
+	RangeBoundDefaultRight = RangeBoundExclusiveRight
 )
 
 func NewRange[T any](lower, upper T) Range[T] {
 	return Range[T]{
 		Lower:      lower,
+		LowerBound: RangeBoundDefaultLeft,
 		Upper:      upper,
-		LowerBound: RangeBoundInclusiveLeft,
-		UpperBound: RangeBoundExclusiveRight,
+		UpperBound: RangeBoundDefaultRight,
 	}
 }
 
-var _ sql.Scanner = (*Range[any])(nil)
-
-func (r *Range[T]) Scan(anySrc any) (err error) {
-	src, ok := anySrc.([]byte)
+func (r *Range[T]) Scan(src any) (err error) {
+	srcB, ok := src.([]byte)
 	if !ok {
-		return fmt.Errorf("pgdialect: Range can't scan %T", anySrc)
+		return fmt.Errorf("unsupported data type: %T", src)
 	}
 
-	if len(src) == 0 {
-		return io.ErrUnexpectedEOF
+	if len(srcB) == 0 {
+		return fmt.Errorf("invalid format: %s", string(srcB))
 	}
-	r.LowerBound = RangeBound(src[0])
-	src = src[1:]
 
-	src, err = scanElem(&r.Lower, src)
-	if err != nil {
+	if string(srcB) == "empty" {
+		return nil
+	}
+
+	// read bounds
+	r.LowerBound = RangeLowerBound(srcB[0])
+	r.UpperBound = RangeUpperBound(srcB[len(srcB)-1])
+	srcB = srcB[1 : len(srcB)-1]
+	if len(srcB) == 0 {
+		return fmt.Errorf("invalid format: %s", string(srcB))
+	}
+
+	l, u, ok := bytes.Cut(srcB, []byte(","))
+	if !ok {
+		return fmt.Errorf("invalid format: %s", string(srcB))
+	}
+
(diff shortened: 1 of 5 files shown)


j2gg0s:
Thank you for your contribution, especially the related test cases.

dialect/pgdialect/range.go
```diff
@@ -1,77 +1,113 @@
...
+var (
+	_ sql.Scanner          = (*MultiRange[any])(nil)
+	_ schema.QueryAppender = (*MultiRange[any])(nil)
+)
+
+func (m *MultiRange[T]) Scan(anySrc any) (err error) {
+	return Array(m).Scan(anySrc)
+}
+
+func (m MultiRange[T]) AppendQuery(fmt schema.Formatter, buf []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
j2gg0s: I think we need to specifically test which form PostgreSQL can accept, as this is a complex and uncommon case.

dialect/pgdialect/range.go
```diff
@@ -1,77 +1,113 @@
...
-type RangeBound byte
+type RangeLowerBound byte
+type RangeUpperBound byte
 
 const (
-	RangeBoundInclusiveLeft  RangeBound = '['
-	RangeBoundInclusiveRight RangeBound = ']'
-	RangeBoundExclusiveLeft  RangeBound = '('
-	RangeBoundExclusiveRight RangeBound = ')'
+	RangeLowerBoundInclusive RangeLowerBound = '['
```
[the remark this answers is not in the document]
thecampagnards: I can keep the same name, but I think it's preferable to keep a different type between lower and upper bound

dialect/pgdialect/range.go
```diff
@@ -1,77 +1,113 @@
...
-	src, err = scanElem(&r.Upper, src)
+	// read range
+	src, err = scanElem(&r.Lower, src)
 	if err != nil {
 		return err
 	}
-
-	if len(src) == 0 {
-		return io.ErrUnexpectedEOF
+	src, err = scanElem(&r.Upper, src)
```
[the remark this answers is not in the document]
thecampagnards: I improved the tests with equality, I also switch to a driver.valuer so we don't need to remove `'`

thecampagnards:
> Thank you for your contribution, especially the related test cases.

Thx for your feedbacks ! With this new updates this should be better for multirange

thecampagnards:
there is also `infinity` bounds, how can we handle it ? Maybe we can have `Range.Lower` and `Range.Upper` as a pointer of `T` and when it's nil, it's `infinity` ?

dialect/pgdialect/range.go
```diff
@@ -1,138 +1,156 @@
...
 }
 
-var _ sql.Scanner = (*Range[any])(nil)
-
 func (r *Range[T]) Scan(anySrc any) (err error) {
-	src, ok := anySrc.([]byte)
-	if !ok {
+	var src []byte
+	switch s := anySrc.(type) {
+	case string:
```
j2gg0s, line 79: Scan only parses the data returned by PostgreSQL, 
so in theory, there should be no case where the input is a string.

dialect/pgdialect/range.go
```diff
@@ -1,138 +1,156 @@
...
+		src = []byte(s)
+	case []byte:
+		src = s
+	default:
 		return fmt.Errorf("pgdialect: Range can't scan %T", anySrc)
 	}
 
+	src = bytes.TrimSpace(src)
```
j2gg0s, line 87: By the same logic, PostgreSQL shouldn’t return any extra spaces.
Moreover, I recall that the upstream code calling this method already trims the spaces.

dialect/pgdialect/range.go
```diff
@@ -1,138 +1,156 @@
...
-	src = src[1:]
 
-	src, err = scanElem(&r.Upper, src)
-	if err != nil {
-		return err
-	}
-
-	if len(src) == 0 {
+	l, u, ok := bytes.Cut(src, []byte(","))
+	if !ok {
```
j2gg0s, line 105: It's invalid, not unexpected eof.

dialect/pgdialect/array.go
```diff
@@ -389,10 +389,6 @@ func arrayScanner(typ reflect.Type) schema.ScannerFunc {
 			}
 		}
 
-		if src == nil {
-			return nil
-		}
```
vmihailenco, line 394: This is array, not range. If you think this should be changed, let's do it in a separate PR ensuring that we have a test for a NULL array.

j2gg0s:
I looked into PostgreSQL’s support for ranges in more detail and tested it on the current master branch.

Unfortunately, it seems that master doesn’t actually work correctly at the moment.
I’ve made some minimal changes to get it working properly, #1227 . 

I think:
	•	Both Lower and Upper need to store pointers to properly represent unbounded values.
	•	An additional variable is needed to indicate whether the range is empty.
	
I haven’t tried MultiRange, but I suspect we’ll need to make some modifications there as well.

Ref:
- PostgreSQL's Document, https://www.postgresql.org/docs/current/rangetypes.html

dialect/pgdialect/range.go
```diff
@@ -1,138 +1,156 @@
...
+}
+
+func (m MultiRange[T]) String() string {
+	if len(m) == 0 {
+		return "{}"
+	}
+	var b []byte
+	b = append(b, '{')
+	for _, r := range m {
+		b = append(b, unquote(appendElem(nil, r))...)
```
vmihailenco, line 36: unquote is not the most elegant and performant solution. And it seems to be in other places as well.

vmihailenco:
>Unfortunately, it seems that master doesn’t actually work correctly at the moment.

It works for me :)

I don't mind making it usable in more cases, but let's not make it right:
- no unquote hacks
- no abundant usage of reflect

j2gg0s:
> > Unfortunately, it seems that master doesn’t actually work correctly at the moment.
> 
> It works for me :)
> 
> I don't mind making it usable in more cases, but let's not make it right:
> 
> * no unquote hacks
> * no abundant usage of reflect

I tried some simple examples, but the SQL resulted in a syntax error — am I using it incorrectly?

```
type Reservation struct {
        Room   int
        During pgdialect.Range[int64] `bun:"type:tsrange"`
}

func main() {
        ctx := context.Background()

        dsn := "postgres://postgres:@localhost:5432/postgres?sslmode=disable"
        sqldb := sql.OpenDB(pgdriver.NewConnector(pgdriver.WithDSN(dsn)))

        db.AddQueryHook(bundebug.NewQueryHook(bundebug.WithVerbose(true)))

        if err := db.ResetModel(ctx, (*Reservation)(nil)); err != nil {
                panic(err)
        }

        reservations := []Reservation{
                {Room: 1, During: pgdialect.NewRange[int64](10, 20)},
        }

        if _, err := db.NewInsert().Model(&reservations).Exec(ctx); err != nil {
                panic(err)
        }

        reservations = reservations[:0]
        if err := db.NewSelect().Model(&reservations).Scan(ctx); err != nil {
                panic(err)
        }
        for _, m := range reservations {
                fmt.Println(m.Room, m.During)
        }
}
```

```
[bun]  14:38:08.021   DROP TABLE           14.704ms  DROP TABLE IF EXISTS "reservations" CASCADE
[bun]  14:38:08.025   CREATE TABLE          3.545ms  CREATE TABLE "reservations" ("room" BIGINT, "during" tsrange)
[bun]  14:38:08.025   INSERT                  701µs  INSERT INTO "reservations" ("room", "during") VALUES (1, [10,20))    pgdriver.Error: ERROR: syntax error at or near "[" (SQLSTATE=42601)
panic: ERROR: syntax error at or near "[" (SQLSTATE=42601)
```

The correct SQL should be:
```sql
INSERT INTO "reservations" ("room", "during") VALUES (1, '[10,20)')
```

vmihailenco:
>The correct SQL should be:

Looks broken to me too. I am using MultiRange which does not have this issue.

thecampagnards:
Yes Range doesnt work for me that why I did this pr.
I let you continue your pr https://github.com/uptrace/bun/pull/1227/files#diff-f04647d00636a6b04da69a178e17d296df572dc1b0e3d966d6d8d310320ae474 and close this one ? Or I update this pr ?
I will try to `go get` on your commit and test on my project
Thx !

dialect/pgdialect/range.go
```diff
@@ -1,138 +1,156 @@
...
+	src = src[1 : len(src)-1]
 	if len(src) == 0 {
 		return io.ErrUnexpectedEOF
 	}
-	if ch := src[0]; ch != ',' {
-		return fmt.Errorf("got %q, wanted %q", ch, ',')
-	}
-	src = src[1:]
 
-	src, err = scanElem(&r.Upper, src)
```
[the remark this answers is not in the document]
thecampagnards: yep but scanElem handle only scanner and time, it doesnt works with int ? or ma I wrong ?

j2gg0s:
> Yes Range doesnt work for me that why I did this pr. I let you continue your pr https://github.com/uptrace/bun/pull/1227/files#diff-f04647d00636a6b04da69a178e17d296df572dc1b0e3d966d6d8d310320ae474 and close this one ? Or I update this pr ? I will try to `go get` on your commit and test on my project Thx !

@thecampagnards 
The current implementation seems to have some bugs. 
If you don’t mind, I’ll go ahead and fix AppendQuery first.

PR #1225, opened 2025-06-26
fix: put spaces before negative numbers following minus signs in format.go to resolve CVE-2024-34359

thunderkatz:
put spaces before negative numbers following minus signs in `format.go` to resolve CVE-2024-34359

### DIFF
--- a/driver/pgdriver/format.go
+++ b/driver/pgdriver/format.go
@@ -66,6 +66,11 @@ func appendArg(b []byte, v interface{}) ([]byte, error) {
 	case nil:
 		return append(b, "NULL"...), nil
 	case int64:
+		// To avoid accidental comments which can lead to SQL injection, put a space before
+		// negative numbers immediately following a minus sign.
+		if v < 0 && len(b) > 0 && b[len(b)-1] == '-' {
+			b = append(b, ' ')
+		}
 		return strconv.AppendInt(b, v, 10), nil
 	case float64:
 		switch {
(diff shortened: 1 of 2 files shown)


Aoang:
Thank you for your valuable contribution!

I'm considering a slightly different approach for this.

Our current implementation utilizes a pre-allocated []byte slice. My thought is that we could integrate a check when appending negative numbers: specifically, if the preceding character is a hyphen (-), we would then insert a space.

CC @j2gg0s What are your thoughts on this alternative?

thunderkatz:
> Thank you for your valuable contribution!
> 
> I'm considering a slightly different approach for this.
> 
> Our current implementation utilizes a pre-allocated []byte slice. My thought is that we could integrate a check when appending negative numbers: specifically, if the preceding character is a hyphen (-), we would then insert a space.
> 
> CC @j2gg0s What are your thoughts on this alternative?

I think this should work. I've gone ahead and switched to doing that.

thunderkatz:
> Awesome work, really appreciate you putting this in!
> 
> Just a small thing, it seems like the imported bytes package isn't being used, so you might need to remove that. Once that's done, it should pass CI.

Done. Thanks for you help here.

guths:
Not to be annoying, but do you have a forecast for the release of a tag with this fix?

thunderkatz:
> Not to be annoying, but do you have a forecast for the release of a tag with this fix?

I do not have permissions to merge this PR (or presumably to make a new release), so I'm waiting for someone with sufficient permissions to get this moving.

j2gg0s:
I will try to handle it in the first half of next week.

j2gg0s:
@thunderkatz Thank you for your work.

Everything is okay, except we got a formatting validation error for the commit style.
Here are some [rules](https://github.com/conventional-changelog/commitlint/?tab=readme-ov-file#what-is-commitlint) you can refer to.

---
If you're not familiar with this, I recommend:

* Using `git rebase` to squash the three commits into one.
* Using `git commit --amend` to change the commit message to `fix(pgdriver): add mandatory space before negative numbers`.

thunderkatz:
> @thunderkatz Thank you for your work.
> 
> Everything is okay, except we got a formatting validation error for the commit style. Here are some [rules](https://github.com/conventional-changelog/commitlint/?tab=readme-ov-file#what-is-commitlint) you can refer to.
> 
> If you're not familiar with this, I recommend:
> 
> * Using `git rebase` to squash the three commits into one.
> * Using `git commit --amend` to change the commit message to `fix(pgdriver): add mandatory space before negative numbers`.

Done. Thanks!

PR #1227, opened 2025-06-27
WIPS: example: pg range

j2gg0s:


dialect/pgdialect/range.go
```diff
@@ -35,58 +41,69 @@ func NewRange[T any](lower, upper T) Range[T] {
...
 
+func NewEmptyRange[T any]() Range[T] {
+	return Range[T]{isEmpty: true}
+}
+
 var _ sql.Scanner = (*Range[any])(nil)
 
-func (r *Range[T]) Scan(anySrc any) (err error) {
-	src, ok := anySrc.([]byte)
+func (r *Range[T]) Scan(raw any) (err error) {
```
j2gg0s, line 50: Valid inputs include:
	•	(lower, upper]
	•	(lower, ]
	•	(, upper]
	•	(,]
	•	empty

j2gg0s:
The current implementation doesn’t seem to actually work. Have we ever used this feature in practice before?
@vmihailenco

vmihailenco:
>The current implementation doesn’t seem to actually work. Have we ever used this feature in practice before?

Yes, I am probably the only user. This is working for me right now:

```go
    Intervals pgdialect.MultiRange[int64] `bun:",multirange"`
```

But yeah, I don't have unbounded and empty ranges.

PR #1230, opened 2025-06-29
chore(breaking): rename Listener.Channel to Listener.CreateChannel

kissejau:
*pgdriver.Listener.Channel is not a Getter.
It`s a method that creates wrapped channel and returns channel of it.

More details [HERE](https://github.com/uptrace/bun/discussions/594#discussioncomment-13611029)

p.s. if major changes can't be accepted, i suggest to add godoc at least.


j2gg0s:
I think this rename is meaningful for preventing misuse. 
However, because it breaks some compatibility, I'll merge it on Wednesday if no one has any other concerns.

Would you mind adding "breaking" to the commit message? 
That would make it easier for people to quickly locate it in the git log.
@kissejau

PR #1245, opened 2025-08-03
feat: Now scanBool, scanInt64 and scanUint64 are compatible with int8/int16/int64 to adapt to some database drivers

qingtao:
The driver provided by the database I am currently using returns a scanError when performing Scan in the bun framework. When the golang type is defined as int8, the returned error is similar to "bun: can't scan hidden(int8) into int8", after debugging, the problem occurs in several functions such as scanInt8 and scanInt16 defined in schema.scanners specified as scanInt64, when the current database driver reaches this point, src is int8, and in scanInt64, when asserting src, only int64 is matched. The same applies to scanUint64 and scanBool. Now, to solve this problem, I have forked a repository In addition, for the three functions scanInt64, scanUint64, and scanBool, matching for int8, int16, int32, etc. has been added. Currently, the program is running normally. I wonder if there will be any other issues with this modification. Also, will the main repository accept this modification?

### DIFF
--- a/schema/scan.go
+++ b/schema/scan.go
@@ -137,7 +137,16 @@ func scanBool(dest reflect.Value, src interface{}) error {
 	case bool:
 		dest.SetBool(src)
 		return nil
-	case int8, int16, int32, int64:
+	case int8:
+		dest.SetBool(src != 0)
+		return nil
+	case int16:
+		dest.SetBool(src != 0)
+		return nil
+	case int32:
+		dest.SetBool(src != 0)
+		return nil
+	case int64:
 		dest.SetBool(src != 0)
 		return nil
 	case []byte:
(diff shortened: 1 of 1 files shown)


Aoang:
Thanks for your detailed explanation and the proposed solution. It appears the database driver you are using might be returning types in a way that diverges from the standard expectations of the `database/sql/driver.Driver` interface, especially regarding how column values are converted during the scan process.

We recommend reviewing the `database/sql` package documentation, particularly around `Rows.Scan` and the `database/sql/driver.ValueConverter` interface. Understanding these core Go features might shed more light on the driver's behavior.

If modifying the driver itself isn't feasible, you could consider implementing a 'driver shim' to intercept and adapt the types returned by your driver before `bun` processes them. Another advanced option might involve creating a custom `bun` dialect to handle these specific type conversions at the ORM level.

While we appreciate your contribution, please be aware that we may not necessarily accept this specific change into the main `bun` project. Our design might rely on certain standard behaviors from database drivers, and such a modification could introduce complexities or deviate from the intended architectural approach.

PR #1256, opened 2025-08-12
fix(db): avoid data racing when cloning db

### LINKED ISSUE #1254: Data race issue when using WithNamedArg
In WithNamedArg source code:

```go
func (db *DB) clone() *DB {
	clone := *db

	l := len(clone.queryHooks)
	clone.queryHooks = clone.queryHooks[:l:l]

	return &clone
}

func (db *DB) WithNamedArg(name string, value interface{}) *DB {
	clone := db.clone()
	clone.fmter = clone.fmter.WithNamedArg(name, value)
	return clone
}
```

clone() copies DB struct which contains DBStats

db struct: 
```go
type DB struct {
	// Must be a pointer so we copy the whole state, not individual fields.
	*noCopyState

	queryHooks []QueryHook

	fmter schema.Formatter
	stats DBStats
}

type DBStats struct {
	Queries uint32
	Errors  uint32
}
```

But DBStats is meant to be threadsafe and is guarded by atomics in other parts of the code base:
```go
func (db *DB) DBStats() DBStats {
	return DBStats{
		Queries: atomic.LoadUint32(&db.stats.Queries),
		Errors:  atomic.LoadUint32(&db.stats.Errors),
	}
}
```

But the unguarded copy here causes the data race issue.

j2gg0s:
Close #1254

### DIFF
--- a/db.go
+++ b/db.go
@@ -258,6 +258,8 @@ func (db *DB) clone() *DB {
 	l := len(clone.queryHooks)
 	clone.queryHooks = clone.queryHooks[:l:l]
 
+	clone.stats = db.DBStats()
+
 	return &clone
 }
 


sayedmurtaza24:
I think the data race is in line 256 where the clone happens.

Issue #1204, opened 2025-06-05
PostgreSQL AS MATERIALIZED CTE


florentcigolotti:
Hi, 

First, I would like to thank you a lot for the work you put in `bun`, it is awesome. 

Here is the context: I work with bun with PostgreSQL and I am using a lot of CTEs. The query optimizer since PG 12 is inlining (if possible) the JOINS in the main query, and that is creating performance issues because it is mixing joins from main request and CTEs. 
It is possible to prevent the inlining with different approach but the one I am interesting in is to create an optimization barrier by using `AS MATERIALIZED` while defining a new CTE (see [here](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-CTE-MATERIALIZATION)).

I did a simple change in bun to implement a `WithMaterialized` (based on tag `v1.2.8`):

```diff
diff --git a/query_base.go b/query_base.go
index 08ff8e5..3763277 100644
--- a/query_base.go
+++ b/query_base.go
@@ -16,6 +16,13 @@ import (
 	"github.com/uptrace/bun/schema"
 )
 
+type WithType int
+
+const (
+	materialized WithType = iota
+	notMaterialized
+)
+
 const (
 	forceDeleteFlag internal.Flag = 1 << iota
 	deletedFlag
@@ -26,6 +33,7 @@ type withQuery struct {
 	name      string
 	query     schema.QueryAppender
 	recursive bool
+	withType  WithType
 }
 
 // IConn is a common interface for *sql.DB, *sql.Conn, and *sql.Tx.
@@ -254,6 +262,16 @@ func (q *baseQuery) addWith(name string, query schema.QueryAppender, recursive b
 		name:      name,
 		query:     query,
 		recursive: recursive,
+		withType:  notMaterialized,
+	})
+}
+
+func (q *baseQuery) addWithMaterialized(name string, query schema.QueryAppender, recursive bool) {
+	q.with = append(q.with, withQuery{
+		name:      name,
+		query:     query,
+		recursive: recursive,
+		withType:  materialized,
 	})
 }
 
@@ -301,7 +319,11 @@ func (q *baseQuery) appendCTE(
 		b = append(b, ")"...)
 	}
 
-	b = append(b, " AS ("...)
+	if cte.withType == materialized {
+		b = append(b, " AS MATERIALIZED ("...)
+	} else {
+		b = append(b, " AS ("...)
+	}
 
 	b, err = cte.query.AppendQuery(fmter, b)
[...]

PR #1260, opened 2025-08-19
feat: add materialize cte support 

fridolin-koch:
I saw https://github.com/uptrace/bun/issues/1204 and took a crack at implementing it. I went with the option pattern instead, but I'm happy to discuss the API

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -22,11 +22,11 @@ const (
 	allWithDeletedFlag
 )
 
-type withQuery struct {
+type WithQuery struct {
 	name            string
 	query           Query
 	recursive       bool
-	materialize     bool
+	materialized    bool
 	notMaterialized bool
 }
 
@@ -104,7 +104,7 @@ type baseQuery struct {
 	tableModel TableModel
 	table      *schema.Table
 
-	with           []withQuery
+	with           []WithQuery
 	modelTableName schema.QueryWithArgs
 	tables         []schema.QueryWithArgs
 	columns        []schema.QueryWithArgs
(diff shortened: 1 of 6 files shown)


vmihailenco:
If we just follow what we did until now, we should just do this:

```
db.NewSelect().WithRecursive() // already exists
db.NewSelect().WithMaterialized() // new API
db.NewSelect().WithNotMaterialized() // new API
```

@fridolin-koch could you update this PR with the new API?

As an idea we could also do this:

```
subq := db.NewWithQuery(name, query).Recursive()
subq := db.NewWithQuery(name, query).Materialized()
subq := db.NewWithQuery(name, query).NotMaterialized()

db.NewSelect().WithQuery(subq)
```

But TBH this does not look worth it in this  particularcase.

fridolin-koch:
> If we just follow what we did until now, we should just do this:

Yeah, I thought about this, but as far as I understand the [postgres docs](https://www.postgresql.org/docs/current/sql-select.html) you can have both MATERIALIZED and RECURSIVE. I felt adding all combinations as methods would be verbose, hence the options. Should I ignore the case and add the methods like you proposed or add all the combinations?

vmihailenco:
Alright, I did not know that.

Let's try this then, i.e. exporting `withQuery` and adding some modifiers:

```go
subq := db.NewWithQuery(name, query).Recursive()
subq := db.NewWithQuery(name, query).Materialized()
subq := db.NewWithQuery(name, query).NotMaterialized()

db.NewSelect().WithQuery(subq)
```

I don't think options API would work in Bun, because we would end up with too many options. Besides, we're already using fluent API which achieves the same.

fridolin-koch:
@vmihailenco Implemented the fluent API as proposed :)

vmihailenco:
Looks good, thanks. In query_test.go you can find a simple way to write some tests for this, but it looks innocent enough as is.

Issue #1228, opened 2025-06-27
SQL injection: standard_conforming_strings


Aoang:
*   **`standard_conforming_strings`:** This setting, when enabled, removes the need to escape backslashes. It has been the **default behavior since PostgreSQL 9.1** (released in 2011, EOL in 2016). For details, see the PostgreSQL documentation and `postgresqlco.nf` for parameter history.
*   **Vulnerability Condition:** The SQL injection risk primarily exists only if `standard_conforming_strings` is **explicitly set to `off`**.
*   **Industry Practice (e.g., `pgx`):** I've observed that drivers like `pgx` incorporate logic to manage this parameter, and in some cases, even prevent its disabling.
*   **Impact Assessment:** Considering `standard_conforming_strings` has been default-on for over a decade, and older versions that default to `off` are well past their end-of-life (e.g., PG 9.1 predates `jsonb`), it's highly improbable that a modern deployment using Bun would have this setting disabled. The parameter largely serves backward compatibility for very legacy environments. Therefore, I believe the **affected user base for this specific vulnerability is extremely limited.**

For: https://github.com/uptrace/bun/issues/1224#issuecomment-2998527181

---

I'll fix this. A CVE will also be submitted, prompting users to update their Bun dependency to resolve this potential issue.

j2gg0s:
Are we still discussing the injection risk caused by minus signs and negative numbers?

Ref: https://www.sonarsource.com/blog/double-dash-double-trouble-a-subtle-sql-injection-flaw/

Aoang:
> Are we still discussing the injection risk caused by minus signs and negative numbers?
> 
> Ref: https://www.sonarsource.com/blog/double-dash-double-trouble-a-subtle-sql-injection-flaw/

The previous issue regarding minus signs and negative numbers was addressed in https://github.com/uptrace/bun/pull/1225. 

This new issue stems from PostgreSQL's `standard_conforming_strings` setting. If this option is disabled, backslashes would need to be escaped to prevent injection, which is not something we currently do.

namtzigla:
Adding a cross-driver nuance for anyone landing here: bun **does** enforce `standard_conforming_strings=on` (and `client_encoding=UTF8`) — but only in its own `driver/pgdriver` (`checkParameterStatus`/`checkRequiredParameter`, `errRequiresParameter`, bypassable via `WithUnsafeStrings`).

[...]

PR #1263, opened 2025-08-22
fix: sql injection #1228

Aoang:
The pgdriver requires that standard_conforming_strings=on and client_encoding=UTF8. Enforce this requirement to prevent SQL injection vulnerabilities.

### DIFF
--- a/driver/pgdriver/driver.go
+++ b/driver/pgdriver/driver.go
@@ -137,6 +137,13 @@ func newConn(ctx context.Context, conf *Config) (*Conn, error) {
 	}
 
 	for k, v := range conf.ConnParams {
+		switch {
+		case k == "standard_conforming_strings" && (v != nil && v != "on"):
+			return nil, errRequiresParameter
+		case k == "client_encoding" && (v != nil && v != "UTF8"):
+			return nil, errRequiresParameter
+		}
+
 		if v != nil {
 			_, err = cn.ExecContext(ctx, fmt.Sprintf("SET %s TO $1", k), []driver.NamedValue{
 				{Value: v},
--- a/driver/pgdriver/proto.go
+++ b/driver/pgdriver/proto.go
@@ -2,6 +2,7 @@ package pgdriver
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/md5"
 	"crypto/tls"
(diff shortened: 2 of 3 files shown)


Aoang:
The core of this PR is to enforce the use of `standard_conforming_strings=on` and `client_encoding=UTF8` for the `pgdriver`. Based on our analysis and PostgreSQL documentation, these parameters are crucial for ensuring correct string literal handling and preventing certain types of SQL injection, particularly when the database or client driver implicitly or explicitly disables standard-compliant string behavior.

While I've thoroughly reviewed the code changes and logically confirmed their necessity, implementing automated tests for this specific modification presents significant challenges.

It requires very old PostgreSQL versions (e.g., pre-8.2 for standard_conforming_strings=off to be default, or versions where UTF8 isn't standard), some of which might even be considered past the operational compatibility of this pgdriver. Alternatively, it would require a user to deliberately and incorrectly configure a modern PostgreSQL instance (standard_conforming_strings=off or non-UTF8 client encoding being explicitly set in a production environment).

Spin up two specially configured PostgreSQL instances within our GitHub Actions CI environment. The one instance would need standard_conforming_strings=off and the other one would need potentially a non-UTF8 encoding.

This setup significantly increases CI complexity, execution time, and resource consumption for a very edge case.

Is there a better way?

driver/pgdriver/driver.go
```diff
@@ -137,6 +137,13 @@ func newConn(ctx context.Context, conf *Config) (*Conn, error) {
...
 
 	for k, v := range conf.ConnParams {
+		switch {
+		case k == "standard_conforming_strings" && (v != nil && v != "on"):
+			return nil, errRequiresParameter
+		case k == "client_encoding" && (v != nil && v != "UTF8"):
+			return nil, errRequiresParameter
+		}
+
 		if v != nil {
```
vmihailenco, line 147: Let's cleanup this a bit by adding:

```
if v == nil { continue }
```

to the top of the loop.

driver/pgdriver/proto.go
```diff
@@ -1100,3 +1109,32 @@ func appendStmtArg(b []byte, v driver.Value) ([]byte, error) {
 		return nil, fmt.Errorf("pgdriver: unexpected arg: %T", v)
 	}
 }
+
+// checkParameterStatus
+//
+// Requires that `standard_conforming_strings=on` and `client_encoding=UTF8`.
```
vmihailenco, line 1115: Could we add a small comment why? ChatGPT can write this for you...

Do we actually want to also disallow client_encoding=ascii? Not even sure if ascii is a real thing for client_encoding...

vmihailenco:
Ideally, this should be under an opt-out flag, e.g. `WithUnsafeStrings(true)` should disable this behavior...

vmihailenco (approved):

PR #1264, opened 2025-08-25
fix: data race in db clone stats

Aoang:
fix: https://github.com/uptrace/bun/issues/1254

### DIFF
--- a/db.go
+++ b/db.go
@@ -59,7 +59,7 @@ type DB struct {
 	queryHooks []QueryHook
 
 	gen   schema.QueryGen
-	stats DBStats
+	stats *DBStats
 }
 
 // noCopyState contains DB fields that must not be copied on clone(),
@@ -81,7 +81,8 @@ func NewDB(sqldb *sql.DB, dialect schema.Dialect, opts ...DBOption) *DB {
 			DB:      sqldb,
 			dialect: dialect,
 		},
-		gen: schema.NewQueryGen(dialect),
+		gen:   schema.NewQueryGen(dialect),
+		stats: &DBStats{},
 	}
 
 	for _, opt := range opts {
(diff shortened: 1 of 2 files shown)


Aoang:
Hi @sayedmurtaza24 , this PR attempts to fix clone `DB.stats` data race. Your review would be appreciated.

vmihailenco:
@Aoang What about moving DBStats to `noCopyState`? I don't think that always copying the stats is desirable, .e.g. it is not obvious that `db.WithNamedArg()` should create a copy of stats.

Also, if `DBStats` are used for collecting metrics, then copying the stats will requires re-instrumenting the client which is just too complicated.

murtaza-pley:
> @Aoang What about moving DBStats to `noCopyState`? I don't think that always copying the stats is desirable, .e.g. it is not obvious that `db.WithNamedArg()` should create a copy of stats.
> 
> Also, if `DBStats` are used for collecting metrics, then copying the stats will requires re-instrumenting the client which is just too complicated.

while on that, perhaps DBStats could be also refactored to use `atomic.Uint32` instead.

Aoang:
> > @Aoang What about moving DBStats to `noCopyState`? I don't think that always copying the stats is desirable, .e.g. it is not obvious that `db.WithNamedArg()` should create a copy of stats.
> > Also, if `DBStats` are used for collecting metrics, then copying the stats will requires re-instrumenting the client which is just too complicated.
> 
> while on that, perhaps DBStats could be also refactored to use `atomic.Uint32` instead.

This is a good idea, but the `DBStats()` method returns a `DBStats` struct. This means that if we use `atomic.Uint32`, we'd either need to split the data or use a separate struct.

We're just fixing a race error here. Perhaps it's better to stick with the current approach?

j2gg0s:
LGTM

It should be noted that if a DB is cloned using a method like WithNamedArg, 
the two DB instances share the same statistics.
At this point, we only need to monitor a single set of data.

PR #1265, opened 2025-08-26
WIP: test: enable race test

Aoang:


Aoang:
I'm a bit confused. Why did a race condition appear when I submitted https://github.com/uptrace/bun/pull/1146, but I can't reproduce it now?

Issue #1237, opened 2025-07-10
m2m relation with extend panic


gwendalF:
Bun version: v1.2.11
Driver:  pgdriver v1.2.11

I have an issue writing a query with a m2m relation when the member is a struct with the `bun:",extend"` struct tag:

type Wrapper struct {
	Order  `bun:",extend"`
	Number string
}
type Order struct {
	bun.BaseModel
	Id    int    `bun:",pk,autoincrement"`
	Items []Item `bun:"m2m:order_to_items,join:Order=Item"`
}

type Item struct {
	bun.BaseModel
	Id   int `bun:",pk,autoincrement"`
	Name string
}

type OrderToItem struct {
	bun.BaseModel
	OrderId int
	Order   *Order `bun:"rel:belongs-to,join:order_id=id"`
	ItemId  int
	Item    *Item `bun:"rel:belongs-to,join:item_id=id"`
}

	order := model.Wrapper{
		Number: "Invoice-1",
	}
	_, err := d.NewInsert().
		Model(&order).
		Exec(t.Context())
	assert.NoError(err)
	items := []model.Item{
		{Name: "itemA"},
		{Name: "itemB"},
	}
	_, err = d.NewInsert().
		Model(&items).
		Exec(t.Context())
	assert.NoError(err)
	_, err = d.NewInsert().
		Model(&[]model.OrderToItem{
			{
				OrderId: order.Id,
				ItemId:  items[0].Id,
			},
			{
				OrderId: order.Id,
				ItemId:  items[1].Id,
			},
		}).
		Exec(t.Context())
	assert.NoError(err)

	var saved model.Wrapper
	err = d.NewSelect().
		Model(&saved).
		Relation("Items").
		Scan(t.Context())
	assert.NoError(err)

There is a panic `panic: reflect: call of reflect.Value.Int on string Value`
It works if I use the Order type directly instead of Wrapper
Here the stacktrace:
```
panic: reflect: call of reflect.Value.Int on string Value [recovered]
	panic: reflect: call of reflect.Value.Int on string Value

goroutine 8 [running]:
testing.tRunner.func1.2({0xa81040, 0xc00039c648})
	/usr/local/go/src/testing/testing.go:1734 +0x21c
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1737 +0x35e
panic({0xa81040?, 0xc00039c648?})
	/usr/local/go/src/runtime/panic.go:792 +0x132
reflect.Value.Int(...)
[...]

PR #1267, opened 2025-08-29
fix(m2m): remap BasePKs to the owner model table for bun:",extend" support

Tiscs:
## Summary

This PR fixes a panic and join-result parking mismatch when loading many-to-many relations from a struct that embeds another model via bun:",extend". The issue was reported upstream (uptrace/bun#1237).
The fix ensures primary key values are collected from the correct owner struct when “extend” flattens fields, and that result parking keys align with the base model.

## Root Cause

- With bun:",extend", the relation metadata (BasePKs) comes from the embedded type. During m2m relation loading:
    - The join builder collected IN (...) values by mixing a root value from one model with field indices intended for another. Under extend, this can point at the wrong field type, causing reflect
panics.
    - The result “parking” keyed base models using BasePK field indices that didn’t match the owner struct under extend, causing “does not have base model with key …” errors.

## Changes

- relation_join.go:
    - Remap BasePK fields to the base model’s table before collecting values.
    - Use the join model’s view (root + parent index) consistently when appending IN (...) values.
- model_table_m2m.go:
    - Remap BasePK fields to the base model’s table before computing baseValues map keys for parking.
- util.go:
    - Add helper remapFieldsToTable to safely map fields by SQL name into the owner table. Keeps logic DRY and avoids mutating cached schema structures.
- Tests:
    - internal/dbtest/orm_test.go: Add testM2MWithExtend using existing Book/Genre/BookGenre and BookWithCommentCount (bun:",extend") to assert Relation("Genres") works.
    - internal/dbtest/extend_m2m_issue1237_test.go: Minimal regression repro that exercises extend+m2m via existing models and asserts no error.
    - Removed ad‑hoc standalone tests to keep suite focused on existing models.

### DIFF
--- a/model_table_m2m.go
+++ b/model_table_m2m.go
@@ -22,9 +22,11 @@ type m2mModel struct {
 var _ TableModel = (*m2mModel)(nil)
 
 func newM2MModel(j *relationJoin) *m2mModel {
-	baseTable := j.BaseModel.Table()
-	joinModel := j.JoinModel.(*sliceTableModel)
-	baseValues := baseValues(joinModel, j.Relation.BasePKs)
+    baseTable := j.BaseModel.Table()
+    joinModel := j.JoinModel.(*sliceTableModel)
+    // Remap BasePKs to the owner (base) model table to account for bun:",extend".
+    remappedBasePKs := remapFieldsToTable(j.Relation.BasePKs, j.BaseModel.Table())
+    baseValues := baseValues(joinModel, remappedBasePKs)
 	if len(baseValues) == 0 {
 		return nil
 	}
--- a/relation_join.go
+++ b/relation_join.go
@@ -185,6 +185,8 @@ func (j *relationJoin) m2mQuery(q *SelectQuery) *SelectQuery {
 	}
 	q = q.Model(m2mModel)
 
+	// Remap BasePKs to the owner (base) model table to account for bun:",extend".
+	remappedBasePKs := remapFieldsToTable(j.Relation.BasePKs, j.BaseModel.Table())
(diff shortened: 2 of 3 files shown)


Tiscs:
Here is a minimal reproducible code.

``` go
package dbtest_test

import (
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/uptrace/bun"
)

// Repro for uptrace/bun#1237: m2m + bun:",extend" must not panic and should load items.
func TestM2MWithExtend_Repro1237(t *testing.T) {
	db := pg(t)

	type Item struct {
		bun.BaseModel
		Id   int `bun:",pk,autoincrement"`
		Name string
	}
	type Order struct {
		bun.BaseModel
		Id    int    `bun:",pk,autoincrement"`
		Items []Item `bun:"m2m:order_to_items,join:Order=Item"`
	}
	type Wrapper struct {
		Order  `bun:",extend"`
		Number string
	}
	type OrderToItem struct {
		bun.BaseModel
		OrderId int
		Order   *Order `bun:"rel:belongs-to,join:order_id=id"`
		ItemId  int
		Item    *Item `bun:"rel:belongs-to,join:item_id=id"`
	}

	mustResetModel(t, ctx, db, (*Item)(nil), (*OrderToItem)(nil), (*Wrapper)(nil))

	order := Wrapper{Number: "Invoice-1"}
	_, err := db.NewInsert().Model(&order).Exec(ctx)
	require.NoError(t, err)

	var saved Wrapper
	err = db.NewSelect().Model(&saved).Relation("Items").Scan(ctx)
	require.NoError(t, err)
}

```

PR #1273, opened 2025-09-25
feat(migrate): add BeforeMigration and AfterMigration

vmihailenco:


migrate/migration.go
```diff
@@ -36,132 +36,137 @@ func (m Migration) IsApplied() bool {
...
-	for _, q := range queries {
-		_, execErr = idb.ExecContext(ctx, q)
-		if execErr != nil {
-			return execErr
-		}
+			panic("not reached")
+		}()
+
+		execErr = migrator.exec(ctx, idb, migration, queries)
+		return retErr
```
tkrop: I think, the new code is now swallowing `execErr` instead of propagating it as result as before.

PR #1277, opened 2025-09-27
feat(migrate): updated default migration naming format

Catalin4513:


bevzzz:
hi @Catalin4513! Why do you think we should change the naming format? The one you're proposing has the same _1min_ precision.

Catalin4513:
> hi @Catalin4513! Why do you think we should change the naming format? The one you're proposing has the same _1min_ precision.

hi, yes both formats have the same 1-minute precision. The change isn't about precision, but rather about human readability and developer experience.

bevzzz:
I see, thank you for the clarification @Catalin4513. Omitting special characters from timestamps is a well-established formatting convention -- their primary function is providing a stable sorting order for migration files. 

For example, [Atlas](https://atlasgo.io/versioned/diff) and [Prisma](https://www.prisma.io/docs/orm/prisma-migrate/getting-started) (as well as many other ORMs) use this convention.

I'm going to close the PR, because we will not be introducing this change.

PR #1280, opened 2025-10-03
fix: return "custom" for unknown dialects instead of "invalid"

### LINKED ISSUE #1276: Dialect Name.String() method returns "invalid" for custom dialects, should return "custom" instead
### Problem Description

The current implementation of `Name.String()` method in `dialect/dialect.go` has a design issue that affects extensibility for custom dialects.
```go
func (n Name) String() string {
	switch n {
	case PG:
		return "pg"
	case SQLite:
		return "sqlite"
	case MySQL:
		return "mysql"
	case MSSQL:
		return "mssql"
	case Oracle:
		return "oracle"
	default:
		return "invalid"  // ← This is problematic
	}
}
```

When developers want to extend Bun to support new database dialects (which is a legitimate use case), they need to define a new `Name` instance for their custom dialect. However, the current `String()` method will return "invalid" for any custom `Name` values, which is misleading and semantically incorrect.

### Expected Behavior

The method should differentiate between:
- **Invalid/uninitialized dialects** - should return "invalid"
- **Custom/unknown dialects** - should return "custom" or the actual name if possible

### Suggested Fix

```go
func (n Name) String() string {
	switch n {
        case Invalid:
                 return "invalid"
	case PG:
		return "pg"
	case SQLite:
		return "sqlite"
	case MySQL:
		return "mysql"
	case MSSQL:
		return "mssql"
	case Oracle:
		return "oracle"
	default:
		return "custom"  // or consider returning the actual name
	}
}
```

This change would improve the framework's extensibility while maintaining backward compatibility for existing code.

coldsmirk:
Previously, the Name.String() method returned "invalid" for all unknown dialects, which was semantically incorrect for custom dialects. This change distinguishes between truly invalid (uninitialized) dialects and custom (unknown but valid) dialects.

Changes:
- Add explicit case for Invalid dialect to return "invalid"
- Change default case to return "custom" for unknown dialects
- Improves framework extensibility for custom dialect implementations

Fixes #1276

### DIFF
--- a/dialect/dialect.go
+++ b/dialect/dialect.go
@@ -4,6 +4,8 @@ type Name int
 
 func (n Name) String() string {
 	switch n {
+	case Invalid:
+		return "invalid"
 	case PG:
 		return "pg"
 	case SQLite:
@@ -15,7 +17,7 @@ func (n Name) String() string {
 	case Oracle:
 		return "oracle"
 	default:
-		return "invalid"
+		return "custom"
 	}
 }
 


Aoang (approved):
LGTM

Issue #1281, opened 2025-10-03
SQLite: WithRecursive generates invalid SQL with extra parentheses in UNION ALL


ilxqx:
## Description

When using `WithRecursive()` with `UnionAll()` in SQLite, Bun generates invalid SQL that wraps each SELECT statement in the UNION with parentheses, which causes a syntax error in SQLite.

## Expected SQL (correct)
```sql
WITH RECURSIVE "category_tree" AS (
  SELECT ... 
  UNION ALL 
  SELECT ...
)
```

## Actual SQL (incorrect)
```sql
WITH RECURSIVE "category_tree" AS (
  (SELECT ...) 
  UNION ALL 
  (SELECT ...)
)
```

SQLite does not support parentheses around individual SELECT statements within a UNION inside a CTE.

## Reproduction

```go
package main

import (
	"context"
	"database/sql"
	"fmt"
	"log"

	"github.com/uptrace/bun"
	"github.com/uptrace/bun/dialect/sqlitedialect"
	"github.com/uptrace/bun/driver/sqliteshim"
)

type Category struct {
	bun.BaseModel `bun:"test_category,alias:c"`
	ID            int64  `bun:"id,pk,autoincrement"`
	Name          string `bun:"name,notnull"`
	ParentID      *int64 `bun:"parent_id"`
}

func main() {
	ctx := context.Background()
	sqldb, _ := sql.Open(sqliteshim.ShimName, "file::memory:?cache=shared")
	defer sqldb.Close()

	db := bun.NewDB(sqldb, sqlitedialect.New())
	
	_, _ = db.NewCreateTable().Model((*Category)(nil)).Exec(ctx)
	
	var results []Category
	err := db.NewSelect().
		WithRecursive("category_tree", db.NewSelect().
			Model((*Category)(nil)).
			Where("name LIKE ?", "%test%").
			UnionAll(
				db.NewSelect().
					Model((*Category)(nil)).
					Join("JOIN ? AS ?", bun.Ident("category_tree"), bun.Ident("ct")).
					JoinOn("?TableAlias.? = ?", bun.Ident("id"), bun.Ident("ct.parent_id")),
			),
		).
		TableExpr("?", bun.Ident("category_tree")).
		Scan(ctx, &results)

	if err != nil {
		log.Fatal(err) // SQL logic error: near "(": syntax error
	}
}
```

## Error
```
SQL logic error: near "(": syntax error (1)
```

## SQLite Verification
```bash
# With parentheses (fails)
[...]

PR #1291, opened 2025-10-31
fix(query): skip UNION wrapping for SQLite

khostya:
Skip parentheses for SQLite UNION queries. Other dialects unchanged.

Fixes https://github.com/uptrace/bun/issues/1281

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -537,8 +537,8 @@ func (q *SelectQuery) appendQuery(
 		b = append(b, "WITH _count_wrapper AS ("...)
 	}
 
-	hasUnion := len(q.union) > 0
-	wrapUnion := hasUnion && gen.Dialect().Name() != dialect.SQLite
+	// SQLite requires UNION parts to be emitted without extra parentheses.
+	wrapUnion := len(q.union) > 0 && gen.Dialect().Name() != dialect.SQLite
 
 	if wrapUnion {
 		b = append(b, '(')


coldsmirk:
Hi @vmihailenco 👋

Just a friendly ping — this PR is still green (all checks passing, no conflicts) and ready for review.

It fixes a real issue with SQLite recursive CTE (#1281), and I'd love to see it land. Could you take a look when you get a chance?

Thanks for your time!

bevzzz:
hi @khostya, thanks for spotting this and opening a PR. I tried rearranging the query a bit to see what was and wasn't valid syntax, and my feeling is that this PR only addresses the case described in the issue.

More generally, SQLite does not allow:

1. Wrapping any statement within a CTE in parenthesis. 

```sh
# Good
sqlite3 :memory: "WITH (SELECT 1) AS t SELECT * FROM t"

# Bad
sqlite3 :memory: "WITH ((SELECT 1)) AS t SELECT * FROM t"
```

2. Wrapping parts of the UNION statement:

```sh
# Good
sqlite3 :memory: "SELECT 1 UNION ALL SELECT 2"

# Bad
sqlite3 :memory: "(SELECT 1) UNION ALL (SELECT 2)"
```

While this PR covers pt. 1, my understanding is it will still produce invalid SQL when UNION is used in a FROM-clause, because `skipUnionWrapping` is only set to true when inside the CTE.

```sql
SELECT * FROM ((SELECT 1) UNION ALL (SELECT 2));
```

---

Looking at `query_select.go` it seems the only place we add additional parenthesis is around a UNION statement. So solving pt.2 (UNION) is sufficient for resolving pt.1 (CTE), at least for SELECT statements.

IMO a good, general solution would be to change how `bun` appends union'ed queries, so that the wrapping logic is in a single place and easy to reason about. E.g., in pseudocode:

```go
if !sqlite { "(" }
for u in unions {
   b = append(b, u.expr...) // UNION ALL
   if !sqlite { "(" }
   b, err = u.query.AppendQuery(gen, b) // SELECT ...
   if !sqlite { ")" }
}
if !sqlite { ")" }
```

then append the root `SelectQuery` as one of the union'ed queries.


WDYT? Are you happy modifying the PR accordingly and will you have the time to do that?
I could also push my suggestions to your branch if you like.

khostya:
@bevzzz thanks, I've updated the PR to handle UNION wrapping for SQLite more generally. Could you take another look?

PR #1302, opened 2025-11-20
fix: Rare panic in indirectAsKey (model_table_has_many.go) when loading complex models

warkanum:
In model_table_has_many.go, the indirectAsKey method has a nil pointer bug. 
Calling reflect.TypeOf(v).Kind() if v is nil causes a panic.

### DIFF
--- a/model_table_has_many.go
+++ b/model_table_has_many.go
@@ -168,7 +168,7 @@ func indirectAsKey(field reflect.Value) any {
 
 	i := field.Interface()
 	if valuer, ok := i.(driver.Valuer); ok {
-		if v, err := valuer.Value(); err == nil {
+		if v, err := valuer.Value(); err == nil && v != nil {
 			switch reflect.TypeOf(v).Kind() {
 			case reflect.Array, reflect.Chan, reflect.Func,
 				reflect.Map, reflect.Pointer, reflect.Slice, reflect.UnsafePointer:


Aoang:
Hi. It looks like the root cause here is an improper implementation of `driver.Valuer`. Could you elaborate on the circumstances under which this occurs?

Implementing a fix for this in the proposed way could inadvertently obscure issues and hinder the troubleshooting process for other incorrect `driver.Valuer` implementations down the line.

warkanum:
Thanks for the update. Apologies for only responding now. 
I agree that it is better for debugging. However, this happens when using sql.NullString or any of the sql.Null types.

When a string is invalid, a nil is returned. e.g. 
```go

type NullString struct {
	String string
	Valid  bool // Valid is true if String is not NULL
}
func (ns NullString) Value() (driver.Value, error) {
	if !ns.Valid {
		return nil, nil
	}
	return ns.String, nil
}
```

It happens with you have nested .Relation of 3 levels or deeper.

warkanum:
More detail on the example. 
```go

type ModelCoreMastertask struct {
	bun.BaseModel `bun:"table:core.mastertask,alias:mastertask"`

	Ridmastertask sql.NullInt64              `json:"rid_mastertask" bun:"rid_mastertask,type:integer,pk,default:nextval('core.identity_mastertask_rid_mastertask'::regclass),"`
	MAL           []*ModelCoreMastertaskitem `json:"MAL,omitempty" bun:"rel:has-many,join:rid_mastertask=rid_mastertask"`
}

type ModelCoreMastertaskitem struct {
	bun.BaseModel     `bun:"table:core.mastertaskitem,alias:mastertaskitem"`
	Ridmastertaskitem sql.NullInt64                   `json:"rid_mastertaskitem" bun:"rid_mastertaskitem,type:integer,pk,default:nextval('core.identity_mastertaskitem_rid_mastertaskitem'::regclass),"`
	Ridmastertask     sql.NullInt64                   `json:"rid_mastertask" bun:"rid_mastertask,type:integer,"`
	MTEV              []*ModelCoreMastertaskitemevent `json:"MTEV,omitempty" bun:"rel:has-many,join:rid_mastertaskitem=rid_mastertaskitem"`
}

type ModelCoreMastertaskitemevent struct {
	bun.BaseModel `bun:"table:core.mastertaskitemevent,alias:mastertaskitemevent"`

	Ridmastertaskitemevent sql.NullInt64            `json:"rid_mastertaskitemevent" bun:"rid_mastertaskitemevent,type:integer,pk,default:nextval('core.identity_mastertaskitemevent_rid_mastertaskitemevent'::regclass),"`
	Ridmastertaskitem      sql.NullInt64            `json:"rid_mastertaskitem" bun:"rid_mastertaskitem,type:integer,"`
	MAL                    *ModelCoreMastertaskitem `json:"MAL,omitempty" bun:"rel:has-one,join:rid_mastertaskitem=rid_mastertaskitem"`
}

```

When the foreign key is null or there are other sql.Null fields in the related models, that error occurs. 
Nil return by the valuer.Value for sql.Null fields.

Aoang:
Looks good.

You need to rewrite the commit message to meet lint checks. https://github.com/uptrace/bun/actions/runs/20338177597/job/58429992529?pr=1302#step:4:14

warkanum:
Thanks. I updated the commit message. 👍🏻

Issue #1335, opened 2026-02-25
Broken implementation of bun.List for the UUIDs in 1.2.17


bgertsberg11:
The bun.In is marked as deprecated in 1.2.17. There is a huge problem with a new implementation of the List feature.  One of the common types of the SQL parameters is a UUID, which happens to be defined, at least in the github.com/google/uuid package, as type UUID [16]byte. This causes the logic in the appendList to print out bytes individually as (1, 2, ,...) instead of printing a formatted UUID.  

This appears to be a side effect of "smart sharing" of the implementation between List and Tuples. Is this recursion really needed for List? 

While this is fixable for the List by just not doing the recursion, the problem could be even bigger for the Tuples as it is open to interpretation of when a value (slice/array) should be expanded to the list or when it shouldn't.  I wonder how loading of one-to-many relations with a single UUID/composite key(potentially with a UUID) is going to work in such design, if bun were to use Tuples feature internally for this.

Since there is no documentation on this feature it is not clear how it is supposed to work.  Is this the case of 1 loop for the List and the nested loops for the Tuples or author meant something more complicated in mind?

NathanBaulch:
Same issue. It looks like the now deprecated `bun.In` handles this with an [explicit byte slice check](https://github.com/uptrace/bun/blob/3c9f8fb8c4f4087617d1ed54317314745dc88172/schema/append.go#L54), whereas `appendList` does not.

PR #1340, opened 2026-02-28
fix: handle []byte and [N]byte in Tuple, separate List from Tuple imp…

vmihailenco:
…lementation

List now uses appendValues (flat, no recursion) while Tuple uses appendTuples (recursive with byte-type special casing). This prevents []byte and [N]byte (e.g. UUIDs) from being expanded byte-by-byte in tuple expressions.


vmihailenco:
Fixes https://github.com/uptrace/bun/issues/1335

PR #1346, opened 2026-03-18
style: fix code style

CyJaySong:
Complete the generic parameter passing of the `setCommentFromContext` function to solve the false positive issue in IDE code analysis

### DIFF
--- a/migrate/auto.go
+++ b/migrate/auto.go
@@ -175,7 +175,8 @@ func NewAutoMigrator(db *bun.DB, opts ...AutoMigratorOption) (*AutoMigrator, err
 		return nil, err
 	}
 	am.dbInspector = dbInspector
-	am.diffOpts = append(am.diffOpts, withCompareTypeFunc(db.Dialect().(sqlschema.InspectorDialect).CompareType))
+	modelInspector := db.Dialect().(sqlschema.InspectorDialect)
+	am.diffOpts = append(am.diffOpts, withCompareTypeFunc(modelInspector.CompareType))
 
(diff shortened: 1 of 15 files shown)


CyJaySong:
@vmihailenco

CyJaySong:
@Aoang

Aoang:
As far as I know, this is a GoLand issue — it fails to properly infer the type here. I recall someone reported a similar problem on JetBrains YouTrack quite a while ago.
Let's wait for feedback from the other maintainers and see if we need to add a workaround in the code to avoid this IDE bug.

CyJaySong:
Standard coding helps IDEs analyze code efficiently, doesn't it?

CyJaySong:
refresh

CyJaySong:
@Aoang I really hope this PR can be recommended. It's not a destructive change and won't affect any logic

bevzzz:
> Standard coding helps IDEs analyze code efficiently, doesn't it?

As @Aoang has correctly pointed out, this is [a bug in GoLand's LSP](https://youtrack.jetbrains.com/projects/GO/issues/GO-19822/Goland-cant-properly-infer-generic-types). The official `gopls` infers the type correctly and ~~anyone using it will have `setCommentFromContext[*AddColumnQuery]()` flagged with a warning about a redundant generic parameter~~ JetBrains are probably going to fix this soon.

I think `setCommentFromContext[*AddColumnQuery]` is not idiomatic and we shouldn't add that only to fix the shortcoming of a third-party LSP.

Issue #1321, opened 2026-01-05
Adding a query with a WHERE OR on a soft-delete model ignores soft-delete condition


john8329:
Using a model with soft-delete enabled, it seems that doing a `q.WhereOr()` adds the condition at the same level of the implicit soft-delete `WHERE` condition. The generated SQL looks like:

```sql
WHERE (cond1)
  OR (cond2)
  AND "table"."deleted_at" IS NULL
```

AND takes precedence, but OR shows the record anyway if it satisfies cond1, effectively making the deleted_at column ignored. I believe that since it's an implicit behavior, the ORM should wrap the query to make it look like:

```sql
WHERE (
    (cond1)
    OR (cond2)
  )
  AND "table"."deleted_at" IS NULL
```

For now I'm manually wrapping it. It's a subtle bug, and a gray area where responsibility is partially the developer's, but it may show the user deleted records, which is also a security issue in some cases.

john8329:
*sigh* bump.

john8329:
Seems like contributing here is pointless. This bot ruins the issues section.

PR #1347, opened 2026-03-18
fix: wrap WHERE conditions in parentheses when soft-delete AND WhereOr are combined

CyJaySong:
for #1321

When using WhereOr() with a soft-delete model, the AND deleted_at IS NULL condition was only binding to the last OR branch due to SQL operator precedence, allowing soft-deleted records to be returned through other OR branches.

Fix: track whether any WHERE condition uses OR separator via a whereHasOr flag on whereBaseQuery. When soft-delete or whereFields conditions need to be ANDed, wrap user WHERE conditions in parentheses if OR is present.

Before: WHERE (cond1) OR (cond2) AND deleted_at IS NULL
After:  WHERE ((cond1) OR (cond2)) AND deleted_at IS NULL

Also save/restore the whereHasOr flag in WhereGroup to prevent group-internal OR conditions from incorrectly triggering the outer wrapping.

### DIFF
--- a/query_base.go
+++ b/query_base.go
@@ -772,9 +772,13 @@ type whereBaseQuery struct {
 
 	where       []schema.QueryWithSep
 	whereFields []*schema.Field
+	whereHasOr  bool
 }
 
 func (q *whereBaseQuery) addWhere(where schema.QueryWithSep) {
+	if strings.Contains(where.Sep, "OR") {
+		q.whereHasOr = true
+	}
 	q.where = append(q.where, where)
 }
 
@@ -844,10 +848,18 @@ func (q *whereBaseQuery) appendWhere(
 	b = append(b, " WHERE "...)
 	startLen := len(b)
 
+	needsExtraCondition := q.isSoftDelete() || q.whereFields != nil
 	if len(q.where) > 0 {
+		wrapParens := needsExtraCondition && q.whereHasOr
+		if wrapParens {
+			b = append(b, '(')
+		}
 		b, err = appendWhere(gen, b, q.where)
 		if err != nil {
 			return nil, err
+		}
+		if wrapParens {
+			b = append(b, ')')
 		}
 	}
 
--- a/query_base.go
+++ b/query_base.go
@@ -776,7 +776,7 @@ type whereBaseQuery struct {
 }
 
 func (q *whereBaseQuery) addWhere(where schema.QueryWithSep) {
-	if strings.Contains(where.Sep, "OR") {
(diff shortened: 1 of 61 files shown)


CyJaySong:
@vmihailenco

CyJaySong:
@Aoang Please review it

Aoang:
> @Aoang Please review it

I'm sorry for the delayed response. Please understand that the maintainers of this library are not working on it full-time.

I did see this PR during the initial round of review, but for various reasons I didn't provide feedback at the time.

One of the most challenging aspects of maintaining an open-source library is that we design it based on the scenarios we anticipate, while trying to support as many use cases as possible — yet we can never know every user's specific situation.

There's no doubt that this is a bug and it should be fixed. However, the fix involves a breaking change. After users update the dependency, they won't receive any obvious warning or error, which makes it quite risky.

I believe the other maintainers are also carefully evaluating the scope and impact of this PR, and considering whether there's a better, less disruptive solution.

This kind of assessment takes time, and sometimes a bit of luck too, as it significantly reduces the time available for deep consideration.

Thank you for your efforts, and please be patient with us.

CyJaySong:
refresh

bevzzz:
I will add my review later this week @Aoang

CyJaySong:
@bevzzz This PR also needs to be followed up

bevzzz:
will do! sorry, been a bit busy at work

CyJaySong:
@bevzzz I am waiting in agony

CyJaySong:
prolong life

CyJaySong:
prolong life

Aoang (commented):
Thanks for your contribution! The fix looks good overall, but `SelectQuery.Clone()` also needs to copy the newly added `whereHasOr` field.

Currently, cloning a soft-delete query containing `WhereOr()` loses this state and produces SQL with the original precedence issue.

It would also be helpful to add a regression test asserting that a query and its clone generate equivalent SQL.

CyJaySong:
@Aoang

query_base.go
```diff
@@ -772,9 +772,13 @@ type whereBaseQuery struct {
 
 	where       []schema.QueryWithSep
 	whereFields []*schema.Field
+	whereHasOr  bool
 }
 
 func (q *whereBaseQuery) addWhere(where schema.QueryWithSep) {
+	if strings.Contains(where.Sep, "OR") {
```
Aoang: Thanks for your contribution. I just noticed an oversight. Users might use it like this: `Where("id = 1").WhereGroup(" or ", ...)`

CyJaySong:
@Aoang

PR #1348, opened 2026-03-18
refactor:`Count`、`Limit`、`Offset`use int64 param

### LINKED ISSUE #1323: The "SelectQuery.Count" method must return int64, otherwise the BigInt type cannot be output in large projects.
MySQL BigInt = Go int64.
But uptrace/bun outputs int32 in the "SelectQuery.Count" method, which prevents the library from being used with BigInt numbers without local fixes. 

Image

CyJaySong:
Fixed #1323

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -376,7 +376,7 @@ func testSelectCount(t *testing.T, db *bun.DB) {
 
 	count, err := q.Count(ctx)
 	require.NoError(t, err)
-	require.Equal(t, 3, count)
+	require.Equal(t, int64(3), count)
 }
 
 func testSelectMap(t *testing.T, db *bun.DB) {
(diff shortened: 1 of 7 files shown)


cvaize:
This fixes my problem. I replaced the vendor version of the package with the corrected one and there were no problems. 

I am currently using stubs: 
```go
func (r *BunRepo) throwInt32(j int) error {
	if j < 0 || j > 2000000000 {
		// int and int32 limit: 2147483647
		return fmt.Errorf("the Count method returns numbers outside the allowed range. It is recommended to correct the return value to the int64 type: %v", j)
	}

	return nil
}

func (r *BunRepo) throwInt64(j int64) error {
	if j < 0 || j > 2000000000 {
		// int and int32 limit: 2147483647
		return fmt.Errorf("the Count method returns numbers outside the allowed range. It is recommended to correct the return value to the int64 type: %v", j)
	}

	return nil
}
```
. 

When the Pull Request is accepted, I will install the new version without this problem.

cvaize:
@CyJaySong Thank you for your work. We also need to correct the commit message. Otherwise, the commitlint stage ends with an error.

CyJaySong:
@vmihailenco

Aoang (approved):
LGTM! Since this is a breaking change, we should wait for a second maintainer’s approval before merging.

Thanks for the PR!

bevzzz (approved):
The same is true for PG, i.e. `COUNT()` returns a BIGINT.
Us returning `int32` instead is a limitation for any large-enough project, so I think it's a good reason to change the data type here.

Issue #1349, opened 2026-03-19
Query error "there is no parameter $1 (SQLSTATE=42P02)" with pgdriver when there is apostrophe in comment


jhekasoft:
If we use `pgdriver` and have query with comment which containes apostrophe (`'`) like this:

```sql
-- don't dfdfdff
SELECT * FROM google_profiles LIMIT $1;
```

Then we retrive error `there is no parameter $1 (SQLSTATE=42P02)`.

This affects other packages that use the `sql.DB` connection with the `pgdriver`. For example, `riverdb`: https://github.com/riverqueue/river/issues/1162

> The apostrophe in doesn't inside the -- comment triggers the QuotedString parser ([format.go:189-207](https://github.com/uptrace/bun/blob/master/driver/pgdriver/format.go#L189)), which then greedily consumes the rest of the query template looking for a closing `'`.

Example with code:

```golang
	// Using pgdriver (recommended)
	sqldb := sql.OpenDB(pgdriver.NewConnector(
		pgdriver.WithDSN("postgres://user:password@localhost:5432/dbname?sslmode=disable"),
	))
	db := bun.NewDB(sqldb, pgdialect.New())

	query := `
		-- don't dfdfdff
		SELECT * FROM google_profiles LIMIT $1;
`
	result, err := db.DB.Exec(query, 1)
	if err != nil {
		fmt.Println(err)
	}
```

And we have:

```
ERROR: there is no parameter $1 (SQLSTATE=42P02)
````

### Proposal

Fix it in `QuotedString`  ([format.go:189-207](https://github.com/uptrace/bun/blob/master/driver/pgdriver/format.go#L189)).

I wil create PR with fix.

jhekasoft:
PR: https://github.com/uptrace/bun/pull/1350

bevzzz:
@Aoang I saw the PR got merged -- was I right to assume the issue can be closed?

Aoang:
> [@Aoang](https://github.com/Aoang) I saw the PR got merged -- was I right to assume the issue can be closed?

Yes, you're right — sorry for forgetting to close it earlier!

PR #1350, opened 2026-03-19
fix: improve formatQuery in pgdriver for queries with apostrophes in comments

jhekasoft:
Fix for query error "there is no parameter $1 (SQLSTATE=42P02)" with pgdriver when there is apostrophe in comment.

Issue: #1349

### DIFF
--- a/driver/pgdriver/format.go
+++ b/driver/pgdriver/format.go
@@ -54,16 +54,10 @@ func formatQuery(query string, args []driver.NamedValue) (string, error) {
 				dst = append(dst, '\'')
 			}
 		case '-':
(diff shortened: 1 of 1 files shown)


driver/pgdriver/format.go
```diff
@@ -53,6 +53,18 @@ func formatQuery(query string, args []driver.NamedValue) (string, error) {
 			} else {
 				dst = append(dst, '\'')
 			}
+		case '-':
```
Aoang: We can add a `LineComment` method to `*parser` that checks if the next byte is `-` and handles the line comment logic accordingly.

PR #1355, opened 2026-03-20
docs: update examples to use WithQueryHook instead of deprecated AddQueryHook

CyJaySong:
Updated 24 example files to use the new WithQueryHook API:
- basic, cursor-pagination, custom-type, migrate
- model-hooks, multi-tenant, fixture, placeholders
- tx-composition, create-table-index
- pg-range, pg-listen, pg-faceted-search
- rel-belongs-to, rel-has-one, rel-has-many
- rel-has-many-polymorphic, rel-join-condition
- rel-many-to-many, rel-many-to-many-self
- trivial/mssql, trivial/mysql, trivial/postgres, trivial/sqlite
- opentelemetry

### DIFF
--- a/example/basic/main.go
+++ b/example/basic/main.go
@@ -20,11 +20,11 @@ func main() {
 	}
 	sqlite.SetMaxOpenConns(1)
 
-	db := bun.NewDB(sqlite, sqlitedialect.New())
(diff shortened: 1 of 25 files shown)


CyJaySong:
@Aoang

PR #1356, opened 2026-03-21
Restore Migration.Up and Migration.Down signatures

### LINKED ISSUE #1297: Migration Add API Broken
In version 1.2.15 the Migration.Add API has breaking changes that make it no longer usable. 

I have been using migrations in Go like so:
```
ms.Add(migrate.Migration{
		Name:    "20241024-initial-migration",
		Comment: "init first model",
		Up: func(ctx context.Context, db *bun.DB) error {
           ...
        },
        Down: func(ctx context.Context, db *bun.DB) error {
           ...
        },
})

```
Since https://github.com/uptrace/bun/commit/d92e29e459ae2804ad48e1b4f6a8147211a47a57 this is no longer possible due to a private type being used in the definition of the `Up` and `Down` members.

Is there a migration path (pun not intended) from `Add` to `Register` that won't break compatibility with migrations that have already been run?

bevzzz:
Resolves #1297 

In an attempt to support templated SQL migration files #1188 changed the signature for `Migration.Up` and `Migration.Down` fields, breaking a part of the `migrate` package's **public contract**. This had the effect of breaking all external code that relied on `Migrations.Add` API previous to this change.

My understanding is that `Migrations.Add` was never meant to be used outside of the `migrate` package, `Migrations.Register` being the officially [documented](https://bun.uptrace.dev/guide/migrations.html#go-based-migrations) and, I assume, the only intended way for client code to register its Go-based migrations. Based on the reactions to the linked issue, there seems to be a number of users who discovered Add (which is exported) and started using it. 

I propose we revert the breaking changes. Both `migrate.Migration` and `migrate.Migrations.Add` have been a part of the public API for a long time and breaking them isn't right. The reason `internalMigrationFunc` was introduced was to support templated migrations and before/after hooks. For each of these the PR presents alternative, b/c-compatible solutions:

1. Migrator can pass template data via `context`. This makes sense, because it is a piece of data scoped to a single migrate/rollback "request".
2. Before/After hooks are handled in a single place. The code which is calling `Up` and `Down` is responsible for applying the hooks, just like it is responsible for calling `MarkApplied`/`MarkUnapplied`.

I also added the missing tests for these two features, see `testSQLMigrations` and `testHooks`.

### DIFF
--- a/internal/dbtest/migrate_test.go
+++ b/internal/dbtest/migrate_test.go
@@ -58,6 +58,7 @@ func TestMigrate(t *testing.T) {
 		{run: testRunMigration},
 		{run: testRunMigrationAssignsNewGroup},
 		{run: testRunMigrationUpErrorPreservesAppliedState},
+		{run: testHooks},
 	}
 
 	testEachDB(t, func(t *testing.T, dbName string, db *bun.DB) {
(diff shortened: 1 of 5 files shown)


bevzzz:
@vmihailenco if possible, I'd like to include this in the next release, WDYT? Do you agree this is a breaking change worth fixing?

PR #1360, opened 2026-04-01
fix: resolve nested Relation() with circular struct composition

### LINKED ISSUE #1243: Regression in v1.2.X: Nested Relation() fails when using struct composition for shared fields and relations
Hi, I’ve run into a regression after upgrading from Bun v1.1.14 to v1.2.14 involving struct composition for shared fields and relationships.

 I use composition in my entity definitions to embed common fields and relationships. Here's a simplified version of the setup:

### Common fields:
```
type IDField struct {
	ID int64 `bun:",pk,unique,autoincrement"`
}

type AgentRelation struct {
	AgentID int64  `bun:",nullzero"`
	Agent   *Agent `bun:"rel:belongs-to,join:agent_id=id"`
}

type LeadRelation struct {
	LeadID int64 `bun:",nullzero"`
	Lead   *Lead `bun:"rel:belongs-to,join:lead_id=id"`
}

type BusinessRelation struct {
	BusinessID int64     `bun:",nullzero"`
	Business   *Business `bun:"rel:belongs-to,join:business_id=id"`
}
```

### Models:
```
type Agent struct {
	IDField
	BusinessRelation

	Associations []*Link `bun:"rel:has-many,join:id=agent_id"`
}

type Business struct {
	IDField
	LeadRelation

	// other fields
}

type Lead struct {
	IDField
	BusinessRelation

	// other fields
}

type Link struct {
	IDField
	AgentRelation
	LeadRelation

        // other fields
}
```

### Query:
```
var agents []entity.Agent
if err := db.NewSelect().
	Model(&agents).
	Relation("Associations.Lead.Business").
	Scan(ctx); err != nil {
	panic(err)
}
```

### Result: It panics with
`sql: Scan error on column index 6, name "lead__business__id":
bun: Link does not have column "lead__business__id"`

### Possible Cause

The bug appears to be in how Bun v1.2 resolves embedded/composed structs during relation joins — particularly when chaining multiple levels (e.g. Agent → Link → Lead → Business)

**Thanks in advance!**

joaquinhuigomez:
When models use shared embedded relation structs that form circular dependencies (e.g. Business -> LeadRelation -> Lead -> BusinessRelation -> Business), `LookupField` fails for nested columns like `lead__business__id` because `StructMap` entries are missing due to initialization order.

The fix splits struct field registration into a `Placeholder` step (registers the StructMap entry with the table pointer) before triggering recursive `InProgress` initialization. This ensures StructMap entries are visible to types in the same circular dependency chain during their own `processFields`.

Fixes #1243

### DIFF
--- a/schema/issue1243_test.go
+++ b/schema/issue1243_test.go
@@ -1,0 +1,77 @@
+package schema
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Models for TestNestedRelationWithSharedComposition (issue #1243).
+// These must be at package level to allow circular type references
+// between Business and Lead via shared embedded relation structs.
+
+type issue1243IDField struct {
+	ID int64 `bun:",pk,autoincrement"`
+}
+
+type issue1243Business struct {
+	issue1243IDField
+	issue1243LeadRelation
+}
+
+type issue1243BusinessRelation struct {
+	BusinessID int64              `bun:",nullzero"`
+	Business   *issue1243Business `bun:"rel:belongs-to,join:business_id=id"`
+}
+
+type issue1243Lead struct {
+	issue1243IDField
+	issue1243BusinessRelation
+}
+
+type issue1243LeadRelation struct {
+	LeadID int64          `bun:",nullzero"`
+	Lead   *issue1243Lead `bun:"rel:belongs-to,join:lead_id=id"`
+}
+
+type issue1243Agent struct {
+	issue1243IDField
+	issue1243BusinessRelation
+
+	Associations []*issue1243Link `bun:"rel:has-many,join:id=agent_id"`
+}
+
+type issue1243AgentRelation struct {
+	AgentID int64           `bun:",nullzero"`
+	Agent   *issue1243Agent `bun:"rel:belongs-to,join:agent_id=id"`
+}
+
+type issue1243Link struct {
+	issue1243IDField
+	issue1243AgentRelation
+	issue1243LeadRelation
+}
+
+func TestNestedRelationWithSharedComposition(t *testing.T) {
+	dialect := newNopDialect()
+	tables := NewTables(dialect)
+
+	// Initialize Link table, which triggers circular initialization:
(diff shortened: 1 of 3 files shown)


Aoang (approved):
Thank you for this PR! The fix looks solid.

schema/issue1243_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Models for TestNestedRelationWithSharedComposition (issue #1243).
+// These must be at package level to allow circular type references
+// between Business and Lead via shared embedded relation structs.
+
+type issue1243IDField struct {
```
bevzzz: nit: could we drop the `issue1243-` prefix from the struct names? I find that the file name `issue1243_test.go` and the doctring above provide enough context. IMO the prefix creates visual noise and makes the test more difficult to read.

suggestion: declare these types inside of the TestNestedRelationWithSharedComposition body; that way we're not polluting the `schema` namespace.

bevzzz (changes requested):

joaquinhuigomez:
@bevzzz thanks for the review. Pushed:

- Dropped the `issue1243-` prefix from the struct names — agreed it reads cleaner with the file/docstring as context.
- Moved the `initStarted` guard into `Table.init()` so callers don't have to manage it.
- Refactored `InProgress` to delegate to `Placeholder` + `init`, matching the structure you suggested.

`TestNestedRelationWithSharedComposition` still passes.

bevzzz:
great @joaquinhuigomez, thank you! 
you'll need to rewrite the last commit message as it's failing the commit-lint rn, but otherwise we're good to go

```
refactor(schema): address review feedback

[details]
```

bevzzz (approved):

joaquinhuigomez:
Status update: this is ready on my end — commit-lint is fixed and CI is green across build, lint, and commitlint. @bevzzz let me know if anything else is needed before merge.

bevzzz (approved):

bevzzz:
Alight, the pipeline is green and the changes LGTM.
Thanks @joaquinhuigomez and let's merge!

PR #1369, opened 2026-04-12
fix: prevent SQL injection via driver.Valuer error messages

### LINKED ISSUE #1307: Possible SQL Injection via driver.Valuer Error Messages
# Summary

When a `driver.Valuer` implementation returns an error, bun embeds the error message directly into the SQL query string via `dialect.AppendError()`. This can lead to SQL injection if the database has certain custom operators defined.

# Severity

**High** - Allows arbitrary SQL execution under specific but achievable conditions.

# Affected Code

- `dialect/append.go:10` - `AppendError()` function
- `schema/append_value.go:271-274` - `appendDriverValue()` calls `AppendError` when `Value()` returns error

```go
func appendDriverValue(gen QueryGen, b []byte, v reflect.Value) []byte {
    value, err := v.Interface().(driver.Valuer).Value()
    if err != nil {
        return dialect.AppendError(b, err)  // embeds error in SQL
    }
    // ...
}
```

# Reproduction

## 1. Create a type whose Value() method returns an error with crafted message:

```go
type BadValue struct{}

func (b BadValue) Value() (driver.Value, error) {
    // Error message crafted to close the ?!() wrapper and inject SQL
    return nil, errors.New("1)); INSERT INTO pwned VALUES (999); --")
}

type TestRow struct {
    bun.BaseModel `bun:"table:test"`
    ID            string   `bun:"id,pk"`
    Bad           BadValue `bun:"bad_col"`
}
```

## 2. The generated SQL contains the error message:

```go
row := &TestRow{ID: "123", Bad: BadValue{}}
q := db.NewInsert().Model(row)
fmt.Println(q.String())
// Output: INSERT INTO "test" ("id", "bad_col") VALUES ('123', ?!(1)); INSERT INTO pwned VALUES (999); --))
```

## 3. Create a prefix operator in PostgreSQL that makes this valid:

```sql
CREATE FUNCTION int_prefix_passthrough(int) RETURNS int AS $$ SELECT $1; $$ LANGUAGE SQL;
CREATE OPERATOR ?! (RIGHTARG = int, FUNCTION = int_prefix_passthrough);
```

## 4. Execute the query - injection succeeds:

```sql
-- Before: pwned table is empty
INSERT INTO test (id, bad_col) VALUES ('123', ?!(1)); INSERT INTO pwned VALUES (999); --))
-- Result: BOTH statements execute. pwned table now contains 999.
```

# Proof of Concept Output

```
INSERT 0 1
INSERT 0 1
 ?column? | id  
----------+-----
 after:   |   1
 after:   | 999
(2 rows)
```

# Attack Requirements

1. A `driver.Valuer` that can return an error (common for custom types with validation)
2. An attacker who can influence the error message content
3. A PostgreSQL database with a `?!` prefix operator defined (unusual but valid)

While requirement #3 is uncommon, the vulnerability exists because:
- Custom operators are a legitimate PostgreSQL feature
- Extensions could define such operators
- The `?!` syntax was chosen arbitrarily and provides no security guarantee

# Additional Concerns

Even without the injection:
1. **Information Leakage**: Error messages (potentially containing sensitive data) are sent to the database server and may be logged
2. **Poor Error Handling**: Errors from `Value()` should be returned to the caller, not embedded in output

# Suggested Fix

Change `AppenderFunc` signature to return `([]byte, error)` and propagate errors properly:

```go
// Before
type AppenderFunc func(gen QueryGen, b []byte, v reflect.Value) []byte

// After  
type AppenderFunc func(gen QueryGen, b []byte, v reflect.Value) ([]byte, error)
```

Or at minimum, use a format that cannot be valid SQL in any database:

```go
func AppendError(b []byte, err error) []byte {
    // Use format that's invalid in all SQL dialects
    b = append(b, "\x00BUN_ERROR:"...)  // NULL byte is never valid in SQL
    b = append(b, err.Error()...)
    b = append(b, '\x00')
    return b
}
```

# Environment

- bun version: v1.2.16 (also confirmed in v1.1.17)
- Go version: 1.25.4
- Database: PostgreSQL 16

Yanhu007:
## Security Fix

Fixes #1307

### Problem

`AppendError` embeds error messages directly into the SQL query string without any escaping. A `driver.Valuer` implementation can return an error with a crafted message that closes the `?!()` wrapper and injects arbitrary SQL.

### Fix

Wrap the error message in single quotes and escape embedded single quotes within the message. This ensures the error message is always treated as a string literal, preventing SQL injection regardless of the error content.

```
Before: ?!(crafted SQL injection here)
After:  ?!('crafted SQL injection here')
```

### DIFF
--- a/dialect/append.go
+++ b/dialect/append.go
@@ -8,9 +8,23 @@ import (
 )
 
 func AppendError(b []byte, err error) []byte {
-	b = append(b, "?!("...)
-	b = append(b, err.Error()...)
-	b = append(b, ')')
+	b = append(b, "?!('"...)
+	b = appendSanitizedError(b, err.Error())
+	b = append(b, "')"...)
+	return b
+}
+
+// appendSanitizedError escapes single quotes in error messages to prevent
+// SQL injection when driver.Valuer returns an error that gets embedded in
+// the query string.
+func appendSanitizedError(b []byte, s string) []byte {
+	for i := 0; i < len(s); i++ {
+		if s[i] == '\'' {
+			b = append(b, '\'', '\'')
+		} else {
+			b = append(b, s[i])
+		}
+	}
 	return b
 }
 


dialect/append.go
```diff
@@ -8,9 +8,23 @@ import (
 )
 
 func AppendError(b []byte, err error) []byte {
-	b = append(b, "?!("...)
-	b = append(b, err.Error()...)
-	b = append(b, ')')
+	b = append(b, "?!('"...)
+	b = appendSanitizedError(b, err.Error())
+	b = append(b, "')"...)
```
bevzzz, line 13:
```suggestion
	b = append(b, "?!("...)
	b = gen.Dialect().AppendString(b, err.Error())
	b = append(b, ")"...)
```

AppendError should probably accept a `QueryGen` as dependency and use it's AppendString method, which already [handles escaping](https://github.com/uptrace/bun/blob/ea46de8e20c11fd6644f6e1aec63052c0a9451f4/schema/dialect.go#L66) exhaustively.

bevzzz (changes requested):

PR #1378, opened 2026-05-09
feat: no ErrNoRows when scan nil struct ptr

CyJaySong:
### previous writing styles
```
user := new(User)
err := db.NewSelect().Model(usert). Where("id = ?", 1).Scan(ctx)
// If there is no data that meets the criteria，err = sql.ErrNoRows， user != nil
```

###  new feature
```
var user *User
err := db.NewSelect().Model(&usert). Where("id = ?", 1).Scan(ctx)
If there is no data that meets the criteria，err = nil， user = nil
```
Compatible with previous writing styles


CyJaySong:
@Aoang Hi Bro，look here

PR #1379, opened 2026-05-09
feat: no ErrNoRows when scan nil struct ptr and nil map and  nil builtin type ptr

CyJaySong:
### previous writing styles
```
user := new(User)
err := db.NewSelect().Model(usert). Where("id = ?", 1).Scan(ctx)
// If there is no data that meets the criteria，err = sql.ErrNoRows， user != nil
```

###  new feature
```
var user *User
err := db.NewSelect().Model(&usert). Where("id = ?", 1).Scan(ctx)
If there is no data that meets the criteria，err = nil， user = nil

var users map[string]any
err := db.NewSelect().Model((*User)(nil)). Where("id = ?", 1).Scan(ctx, &users)
If there is no data that meets the criteria，err = nil， users = nil

var phone *string
err := db.NewSelect().Model((*User)(nil)). Where("id = ?", 1).Column("phone").Scan(ctx, &phone)
If there is no data that meets the criteria，err = nil， phone = nil
```
Compatible with previous writing styles

### DIFF
--- a/internal/dbtest/db_test.go
+++ b/internal/dbtest/db_test.go
@@ -262,6 +262,7 @@ func TestDB(t *testing.T) {
 		{testSelectMap},
 		{testSelectMapSlice},
 		{testSelectStruct},
+		{testSelectStructNilPtr},
 		{testSelectNestedStructValue},
 		{testSelectNestedStructPtr},
 		{testSelectStructSlice},
@@ -442,6 +443,51 @@ func testSelectStruct(t *testing.T, db *bun.DB) {
 	err = db.NewSelect().ColumnExpr("1 as unknown_column").Scan(ctx, model)
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "Model does not have column")
+}
+
+func testSelectStructNilPtr(t *testing.T, db *bun.DB) {
+	type Model struct {
+		Num int
+		Str string
+	}
+
+	// Scan a row into a nil **Model: the inner pointer should be allocated and populated.
+	var model *Model
+	err := db.NewSelect().
+		ColumnExpr("10 AS num, 'hello' AS str").
+		Scan(ctx, &model)
+	require.NoError(t, err)
+	require.NotNil(t, model)
+	require.Equal(t, 10, model.Num)
+	require.Equal(t, "hello", model.Str)
+
+	// No rows + nil **Model: should NOT return ErrNoRows; inner pointer stays nil.
+	var empty *Model
+	err = db.NewSelect().
+		TableExpr("(SELECT 42 AS num) AS t").
+		Where("1 = 2").
+		Scan(ctx, &empty)
+	require.NoError(t, err)
+	require.Nil(t, empty)
+
+	// Regression: a non-nil *Model (existing behaviour) must still return ErrNoRows.
+	nonNil := new(Model)
+	err = db.NewSelect().
+		TableExpr("(SELECT 42 AS num) AS t").
+		Where("1 = 2").
+		Scan(ctx, nonNil)
+	require.Equal(t, sql.ErrNoRows, err)
+
+	// Scan a row into a non-nil **Model: the existing inner pointer is reused.
+	existing := &Model{}
+	preserved := existing
+	err = db.NewSelect().
+		ColumnExpr("7 AS num, 'world' AS str").
+		Scan(ctx, &existing)
+	require.NoError(t, err)
+	require.Same(t, preserved, existing)
+	require.Equal(t, 7, existing.Num)
+	require.Equal(t, "world", existing.Str)
 }
 
 func testSelectNestedStructValue(t *testing.T, db *bun.DB) {
--- a/model.go
+++ b/model.go
@@ -124,6 +124,10 @@ func _newModel(db *DB, dest any, scan bool) (Model, error) {
 		return newMapModel(db, mapPtr), nil
 	case reflect.Struct:
 		return newStructTableModelValue(db, dest, v), nil
+	case reflect.Pointer:
+		if v.Type().Elem().Kind() == reflect.Struct {
+			return newStructTableModelValue(db, dest, v), nil
+		}
 	case reflect.Slice:
 		switch elemType := sliceElemType(v); elemType.Kind() {
 		case reflect.Struct:
@@ -200,8 +204,7 @@ func validMap(typ reflect.Type) error {
 
 func isSingleRowModel(m Model) bool {
 	switch m.(type) {
-	case *mapModel,
-		*structTableModel,
+	case *structTableModel,
 		*scanModel:
 		return true
 	default:
--- a/model.go
+++ b/model.go
@@ -204,10 +204,17 @@ func validMap(typ reflect.Type) error {
 
 func isSingleRowModel(m Model) bool {
 	switch m.(type) {
-	case *structTableModel,
+	case *mapModel,
+		*structTableModel,
 		*scanModel:
 		return true
 	default:
 		return false
 	}
+}
+
+// nilModel is implemented by models whose destination may be a nil pointer
+// that should not be treated as "no rows" when an empty result is scanned.
+type nilModel interface {
+	isNil() bool
 }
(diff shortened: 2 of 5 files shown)


CyJaySong:
@Aoang Hi Bro，look here

query_base.go
```diff
@@ -630,7 +631,24 @@ func (q *baseQuery) _scan(
 	}
 
 	if numRow == 0 && hasDest && isSingleRowModel(model) {
```
bevzzz, line 632: suggestion: could we instead apply this logic to how `hasDest` is calculated?
If I understand your idea correctly, it is to say that a `nil` dest is the same as "no dest". If so, then `hasDest` should be `false` for such a case and then the if-statement won't fire.

bevzzz (changes requested):

query_base.go
CyJaySong: I changed a way of judgment

model.go
```diff
@@ -200,8 +204,7 @@ func validMap(typ reflect.Type) error {
 
 func isSingleRowModel(m Model) bool {
 	switch m.(type) {
-	case *mapModel,
-		*structTableModel,
+	case *structTableModel,
```
bevzzz: question: iiuc `*mapModel` is not considered "single row model", because the return value is `false` -- why is that?

query_base.go
```diff
@@ -630,6 +630,9 @@ func (q *baseQuery) _scan(
...
+		if nm, ok := model.(*structTableModel); ok && nm.isNil() {
```
bevzzz: question: shouldn't the fix you propose also hold for a nil map?

```go
var nilMap map[string]any
err := selectQuery.Scan(ctx, &nilMap)
require.NoError(err) // SQL_NO_ROWS, but the dest is nil, so no error
```

CyJaySong: yes

model.go
CyJaySong: for a nil map

model.go
```diff
@@ -208,3 +212,9 @@ func isSingleRowModel(m Model) bool {
 		return false
 	}
 }
+
+// nilModel is implemented by models whose destination may be a nil pointer
+// that should not be treated as "no rows" when an empty result is scanned.
+type nilModel interface {
+	isNil() bool
+}
```
bevzzz, line 241: question: does `scanModel` also need to implement `nilModel`?

I am not too familiar with this area of the code but I'm wondering if below is another case where we _should not_ return ErrNoRows:

```go
var nilInt *int
db.NewSelect().
  TableExpr("SELECT 1").Where("FALSE").
  Scan(ctx, &nilInt)
```

Lmk if the question makes sense

CyJaySong: ok，supported

CyJaySong:
@bevzzz

bevzzz:
@CyJaySong I'll take another pass at the PR tomorrow, just to make sure I understand the changes well. Thanks for your patience and the effort you're putting in contributing to Bun.

CyJaySong:
@bevzzz Thanks for your recognition. I‘m looking forward to a smooth merger

CyJaySong:
@bevzzz I am waiting in agony

CyJaySong:
prolong life

CyJaySong:
prolong life

internal/dbtest/db_test.go
```diff
@@ -444,6 +445,51 @@ func testSelectStruct(t *testing.T, db *bun.DB) {
...
+	preserved := existing
+	err = db.NewSelect().
+		ColumnExpr("7 AS num, 'world' AS str").
+		Scan(ctx, &existing)
+	require.NoError(t, err)
+	require.Same(t, preserved, existing)
+	require.Equal(t, 7, existing.Num)
+	require.Equal(t, "world", existing.Str)
+}
+
```
Aoang, line 492:
```go
func testPR1379(t *testing.T, db *bun.DB) {
	t.Run("pointer to time.Time", func(t *testing.T) {
		var got *time.Time
		err := db.NewSelect().
			ColumnExpr("TIMESTAMPTZ '2026-01-02 03:04:05+00'").
			Scan(ctx, &got)

		require.NoError(t, err)
		require.NotNil(t, got)
		require.Equal(t, time.Date(2026, 1, 2, 3, 4, 5, 0, time.UTC), got.UTC())
	})

	t.Run("pointer to sql.Scanner struct", func(t *testing.T) {
		var got *sql.NullString
		err := db.NewSelect().
			ColumnExpr("'hello'::text").
			Scan(ctx, &got)

		require.NoError(t, err)
		require.NotNil(t, got)
		require.True(t, got.Valid)
		require.Equal(t, "hello", got.String)
	})
}
```

The current implementation introduces a backward compatibility regression.

Aoang (commented):
Thanks for your contribution, and sorry for the long review time!

So far, I've spotted two breaking changes, and it's still unclear if there are others.

Like `if err := query.Scan(ctx, &id); errors.Is(err, sql.ErrNoRows) {}`, upgrading will break application logic since sql.ErrNoRows won't be triggered anymore. This is a tough breaking change to introduce. At a minimum, this feature should be optional/opt-in, or exposed via a separate API like ScanSilent.

To push this forward, I suggest we discuss the design first in an issue to find a reasonable solution. Meanwhile, this PR remains very valuable, especially as it points out gaps in our regression test suite.

CyJaySong:
@Aoang

Aoang:
Thanks for the update. I retested the latest commit against PostgreSQL. The main backward-compatibility concern is still unresolved. Existing applications may already rely on `sql.ErrNoRows`.

```go
func testPR1379ErrNoRows(t *testing.T, db *bun.DB) {
	t.Run("nil pointer destination", func(t *testing.T) {
		var got *int
		err := db.NewSelect().
			ColumnExpr("1").
			Where("FALSE").
			Scan(ctx, &got)

		require.ErrorIs(t, err, sql.ErrNoRows)
		require.Nil(t, got)
	})

	t.Run("allocated pointer destination", func(t *testing.T) {
		got := new(int)
		err := db.NewSelect().
			ColumnExpr("1").
			Where("FALSE").
			Scan(ctx, &got)

		require.ErrorIs(t, err, sql.ErrNoRows)
		require.NotNil(t, got)
	})
}
```

CyJaySong:
> Thanks for the update. I retested the latest commit against PostgreSQL. The main backward-compatibility concern is still unresolved. Existing applications may already rely on `sql.ErrNoRows`.
> 
> ```go
> func testPR1379ErrNoRows(t *testing.T, db *bun.DB) {
> 	t.Run("nil pointer destination", func(t *testing.T) {
> 		var got *int
> 		err := db.NewSelect().
> 			ColumnExpr("1").
> 			Where("FALSE").
> 			Scan(ctx, &got)
> 
> 		require.ErrorIs(t, err, sql.ErrNoRows)
> 		require.Nil(t, got)
> 	})
> 
> 	t.Run("allocated pointer destination", func(t *testing.T) {
> 		got := new(int)
> 		err := db.NewSelect().
> 			ColumnExpr("1").
> 			Where("FALSE").
> 			Scan(ctx, &got)
> 
> 		require.ErrorIs(t, err, sql.ErrNoRows)
> 		require.NotNil(t, got)
> 	})
> }
> ```

@Aoang  Oh, that was by design, not a bug. It fixes the inconsistent no raws handling behavior. I think this counts as a breaking change.

CyJaySong:
@Aoang Hi, Bro, do you agree with my idea?

Aoang:
> @Aoang Hi, Bro, do you agree with my idea?

Thanks for your contribution! There's no doubt that your implementation provides a much more elegant and reasonable way to handle things.

Looking at the evolution of Go ORM ecosystems—from early days like xorm and gorm to tools like sqlx—none of them handled sql.ErrNoRows as cleanly as this approach does.

My main concern, however, is that this is a breaking change that might force users to make extensive refactoring when upgrading to the new version. I did a quick survey of how users currently write code with bun: while most codebases won't be affected, there are still instances where incorrect usages slipped through due to a lack of code review (or oversight during review). These usages work fine in the current version, but this change would break them.

Let's wait for feedback from other maintainers to see what they think.

CyJaySong:
prolong life

PR #1391, opened 2026-07-17
Add slice of byte array support to pgdialect

NathanBaulch:
I need to store an array of [ULIDs](https://github.com/oklog/ulid) in a `bytea[]` field. pgdialect correctly identifies this type as an array of `[16]byte` but generates an invalid SQL literal due to single quotes within a single quoted string, eg `'{'\x010203','\x040506'}'`.

This is a very simple fix with unit tests.


NathanBaulch:
I'm calling this a `fix` rather than a `feat` because it generates invalid SQL.

Aoang (approved):

PR #1392, opened 2026-07-20
fix: copy execution state in SelectQuery.Clone

### LINKED ISSUE #1388: SelectQuery.Clone drops execution state (Conn, Err, soft-delete flags, WherePK, CTE materialization)
## Summary

`(*bun.SelectQuery).Clone` is documented as creating a deep copy, but in v1.2.18
and current `master` (`21dd5513a6697aa1d2d87e23155727944fdde198`) it does not copy several
fields that affect query semantics or execution routing:

- `baseQuery.conn`
- `baseQuery.err`
- `baseQuery.flags`
- `whereBaseQuery.whereFields`
- `WithQuery.materialized`
- `WithQuery.notMaterialized`

This is more than clone isolation: executing the clone can target a different
connection, lose a builder error or soft-delete policy, turn `WherePK` into an
unfiltered query, and drop PostgreSQL CTE materialization hints.

This is related to #1298, which fixed the separate empty-but-non-nil
`distinctOn` clone loss. The fields below are still omitted after that fix.

## Current implementation

The `SelectQuery` literal copies `db`, `table`, `model`, `tableModel`, and the
public builder slices, but leaves `conn`, `err`, `flags`, and `whereFields` at
their zero values. The `WithQuery` loop copies `name`, `recursive`, and `query`,
but not its two materialization booleans.

Relevant current-master code:

https://github.com/uptrace/bun/blob/21dd5513a6697aa1d2d87e23155727944fdde198/query_select.go#L1175-L1292

The omitted fields are defined here:

https://github.com/uptrace/bun/blob/21dd5513a6697aa1d2d87e23155727944fdde198/query_base.go#L26-L32

https://github.com/uptrace/bun/blob/21dd5513a6697aa1d2d87e23155727944fdde198/query_base.go#L98-L114

https://github.com/uptrace/bun/blob/21dd5513a6697aa1d2d87e23155727944fdde198/query_base.go#L770-L775

## Reproduction cases

These cases reproduce independently:

```go
// 1. WherePK: original emits a PK predicate; Clone does not.
original := db.NewSelect().Model(&User{ID: 42}).WherePK()
cloned := original.Clone()

// 2. Builder error: Clone returns a query with Err() == nil.
sentinel := errors.New("sentinel")
cloned = db.NewSelect().Err(sentinel).Clone()

// 3. Explicit connection: Clone resolves through the base DB/resolver instead
// of the connection supplied to Conn.
cloned = db.NewSelect().Conn(explicitConn).Clone()

// 4. Soft-delete flags: WhereDeleted / WhereAllWithDeleted policy is reset.
cloned = db.NewSelect().Model((*User)(nil)).WhereAllWithDeleted().Clone()

// 5. CTE hints: AS MATERIALIZED / AS NOT MATERIALIZED disappears.
with := bun.NewWithQuery(
    "cte",
    db.NewSelect().ColumnExpr("1 AS n"),
).Materialized()
cloned = db.NewSelect().WithQuery(with).TableExpr("cte").Clone()
```

The most dangerous case is `WherePK`: a caller can clone a query specifically
to execute or further refine it and receive an unfiltered statement without an
error.

## Expected behavior

The clone should preserve all state that changes the original query's SQL,
error result, or execution connection. Top-level builder slices should remain
independent as they are today; referenced model values and nested query objects
can keep the existing sharing policy if recursive clone semantics are out of
scope.

## Suggested narrow fix

- Copy `conn`, `err`, and `flags` into the cloned `baseQuery`.
- Clone the `whereFields` slice (the `*schema.Field` values can be shared).
- Copy `materialized` and `notMaterialized` in each `WithQuery` entry.
- Add focused regression tests for each field.

I can prepare a PR for this narrow change if the expected clone contract above
matches the maintainers' intent.

Yusufihsangorgel:
Closes #1388

`SelectQuery.Clone` copies the builder slices but leaves several execution-affecting fields at their zero values:

- `baseQuery.conn`: the clone targets the default pool, not the original's transaction or connection.
- `baseQuery.err`: a recorded builder error is dropped, so the clone runs instead of returning it.
- `baseQuery.flags`: `WhereDeleted` / `WhereAllWithDeleted` are lost.
- `whereBaseQuery.whereFields`: `WherePK` becomes an unfiltered scan.
- `WithQuery.materialized` / `notMaterialized`: `MATERIALIZED` hints disappear from the CTE.

I copy these fields, including the two `WithQuery` hints. `whereFields` goes through the same nil check `cloneArgs` uses, so a non-nil empty slice survives and a nil one stays nil.

The execution-state test fails on master and passes here. `go test ./...` passes in the root module.

### DIFF
--- a/query_select.go
+++ b/query_select.go
@@ -1193,6 +1193,14 @@ func (q *SelectQuery) Clone() *SelectQuery {
 		copy(clone, args)
 		return clone
 	}
+	cloneWhereFields := func(fields []*schema.Field) []*schema.Field {
+		if fields == nil {
+			return nil
+		}
+		clone := make([]*schema.Field, len(fields))
+		copy(clone, fields)
+		return clone
+	}
 	cloneHints := func(hints *indexHints) *indexHints {
 		if hints == nil {
 			return nil
@@ -1213,15 +1221,19 @@ func (q *SelectQuery) Clone() *SelectQuery {
 		whereBaseQuery: whereBaseQuery{
 			baseQuery: baseQuery{
 				db:             q.db,
+				conn:           q.conn,
 				table:          q.table,
 				model:          q.model,
+				err:            q.err,
 				tableModel:     tableModel,
 				with:           make([]WithQuery, len(q.with)),
 				tables:         cloneArgs(q.tables),
 				columns:        cloneArgs(q.columns),
 				modelTableName: q.modelTableName,
+				flags:          q.flags,
 			},
-			where: make([]schema.QueryWithSep, len(q.where)),
+			where:       make([]schema.QueryWithSep, len(q.where)),
+			whereFields: cloneWhereFields(q.whereFields),
 		},
 
 		idxHintsQuery: idxHintsQuery{
@@ -1246,9 +1258,11 @@ func (q *SelectQuery) Clone() *SelectQuery {
 
 	for i, w := range q.with {
 		clone.with[i] = WithQuery{
-			name:      w.name,
-			recursive: w.recursive,
-			query:     w.query, // TODO: maybe clone is need
+			name:            w.name,
+			recursive:       w.recursive,
+			materialized:    w.materialized,
+			notMaterialized: w.notMaterialized,
+			query:           w.query, // TODO: maybe clone is need
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Aoang (approved):
LGTM

PR #1393, opened 2026-07-20
fix: return an error instead of panicking when scanning JSON into an unaddressable value

### LINKED ISSUE #1306: Panic in scanJSON causes deadlock
Go version: go1.25.0 darwin/arm64.
Bun version: v1.2.16
PostgreSQL version: 17

I worked with a nullable JSONB column in PostgreSQL and found a case where an uninitialized empty interface in the model definition causes the `scanJSON` ([code](https://github.com/uptrace/bun/blob/c951ef4897c54e53a3aeb112f237e1b82a284755/schema/scan.go#L361)) function to panic. In this case , the`Exec` and `Scan` methods stay blocked and the database connection has `idle in transaction` state.

In the code snippet below, I want to insert two rows: one with JSONB data and one without (expecting NULL in that column). The `nullzero` annotation forces query executor to select data from the table (query will have `RETURNING data` statement). We can see that the `Data` field of the second slice element is uninitialized, so it is empty interface. Empty interface is a unaddressable value, so in `scanJSON` function we can not call `dest.Addr().Interface()` for it.

```go
package main

import (
	"context"
	"database/sql"
	"log"
	"time"

	"github.com/uptrace/bun"
	"github.com/uptrace/bun/dialect/pgdialect"
	"github.com/uptrace/bun/driver/pgdriver"
)

func main() {
	dsn := "postgres://postgres:@localhost:55432/postgres?sslmode=disable"
	sqldb := sql.OpenDB(pgdriver.NewConnector(pgdriver.WithDSN(dsn)))

	log.Println("Connected to database")

	_, err := db.Exec("CREATE TABLE IF NOT EXISTS test (id BIGINT PRIMARY KEY, data JSONB); TRUNCATE test;")
	if err != nil {
		log.Fatalln(err)
	}

	log.Println("Created \"test\" table")

	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer cancel()

	rows := []model{
		{
			ID: 1,
			Data: map[string]any{
				"foo": "bar",
			},
		},
		{
			ID: 2,
		},
	}

	_, err = db.NewInsert().
		Model(&rows).
		Exec(ctx)
	if err != nil {
		log.Fatalln(err)
	}

	log.Println(rows)
}

type model struct {
	bun.BaseModel `bun:"table:test"`

	ID   int64 `bun:"id,pk"`
	Data any   `bun:"data,nullzero,type:jsonb"`
}
```

Stack trace:

```
2025/11/25 21:52:20 Connected to database
2025/11/25 21:52:20 Created "test" table
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [sync.RWMutex.Lock]:
sync.runtime_SemacquireRWMutex(0x10?, 0x60?, 0x14000002301?)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/runtime/sema.go:105 +0x28
sync.(*RWMutex).Lock(0x104b54a78?)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/sync/rwmutex.go:155 +0xf4
database/sql.(*Rows).close(0x1400011a280, {0x0, 0x0})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3443 +0x80
database/sql.(*Rows).Close(0x1400011a280)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3439 +0x30
panic({0x1047ad6c0?, 0x10481aef0?})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/runtime/panic.go:783 +0x120
reflect.Value.Addr({0x1047ca980?, 0x1400011e1b0?, 0x104819df0?})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/reflect/value.go:268 +0x64
github.com/uptrace/bun/schema.scanJSON({0x1047ca980?, 0x1400011e1b0?, 0x104b54a78?}, {0x1047a9500?, 0x14000138198?})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/schema/scan.go:361 +0x78
github.com/uptrace/bun/schema.scanJSONIntoInterface({0x1047badc0?, 0x1400011e1e8?, 0x1047e6620?}, {0x1047a9500, 0x14000138198})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/schema/scan.go:535 +0x140
github.com/uptrace/bun/schema.(*Field).ScanWithCheck(0x1047e6620?, {0x1047badc0?, 0x1400011e1e8?, 0x104780f70?}, {0x1047a9500?, 0x14000138198?})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/schema/field.go:132 +0x88
github.com/uptrace/bun/schema.(*Field).ScanValue(0x14000142300, {0x1047e6620?, 0x1400011e1e0?, 0x0?}, {0x1047a9500, 0x14000138198})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/schema/field.go:125 +0xc4
github.com/uptrace/bun.(*structTableModel).scanColumn(0x14000150000, {0x14000154000, 0x4}, {0x1047a9500, 0x14000138198})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/model_table_struct.go:314 +0xc0
github.com/uptrace/bun.(*structTableModel).ScanColumn(0x14000150000, {0x14000154000, 0x4}, {0x1047a9500?, 0x14000138198?})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/model_table_struct.go:294 +0x30
github.com/uptrace/bun.(*structTableModel).Scan(0x104a13430?, {0x1047a9500?, 0x14000138198?})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/model_table_struct.go:290 +0x98
database/sql.convertAssignRows({0x10480b380, 0x14000150000}, {0x1047a9500, 0x14000138198}, 0x1400011a280)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/convert.go:394 +0x1a48
database/sql.(*Rows).scanLocked(0x1400011a280, {0x14000023bf8, 0x1, 0x14000023b28?})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3399 +0x1f0
database/sql.(*Rows).Scan(0x1400011a280, {0x14000023bf8, 0x1, 0x1400011a280?})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3374 +0x9c
github.com/uptrace/bun.(*structTableModel).scanRow(0x14000150000, {0x10481eeb0, 0x1400013c000}, 0x1400011a280, {0x14000023bf8, 0x1, 0x1})
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/model_table_struct.go:275 +0x58
github.com/uptrace/bun.(*sliceTableModel).ScanRows(0x14000150000, {0x10481eeb0, 0x1400013c000}, 0x1400011a280)
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/model_table_slice.go:76 +0x234
github.com/uptrace/bun.(*baseQuery)._scan(0x1400000c288?, {0x10481eeb0, 0x1400013c000}, {0x10481ef90?, 0x14000140000?}, {0x14000152000, 0x5c}, {0x10481e1e0, 0x14000150000}, 0x0)
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/query_base.go:622 +0xbc
github.com/uptrace/bun.(*baseQuery).scan(0x14000140000, {0x10481eeb0?, 0x1400013c000?}, {0x10481ef90, 0x14000140000}, {0x14000152000, 0x5c}, {0x10481e1e0, 0x14000150000}, 0x0)
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/query_base.go:604 +0xa0
github.com/uptrace/bun.(*InsertQuery).scanOrExec(0x14000140000, {0x10481eeb0, 0x1400013c000}, {0x0, 0x0, 0x0}, 0x0)
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/query_insert.go:614 +0x28c
github.com/uptrace/bun.(*InsertQuery).Exec(...)
	/Users/gopher/go/pkg/mod/github.com/uptrace/bun@v1.2.16/query_insert.go:569
main.main()
	/Users/gopher/Projects/sql-db-issue/bun/main.go:46 +0x318

goroutine 34 [sync.Mutex.Lock]:
internal/sync.runtime_SemacquireMutex(0x104a23700?, 0x50?, 0x14000049e18?)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/runtime/sema.go:95 +0x28
internal/sync.(*Mutex).lockSlow(0x1400011a2b8)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/internal/sync/mutex.go:149 +0x170
internal/sync.(*Mutex).Lock(...)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/internal/sync/mutex.go:70
sync.(*Mutex).Lock(...)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/sync/mutex.go:46
sync.(*RWMutex).Lock(0x1400011a2b8)
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/sync/rwmutex.go:150 +0x6c
database/sql.(*Rows).close(0x1400011a280, {0x10481d000, 0x104a49940})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3443 +0x80
database/sql.(*Rows).awaitDone(0x1400011a280, {0x10481eeb0, 0x1400013c000}, {0x0, 0x0}, {0x10481ee40, 0x14000158050})
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:3020 +0x19c
created by database/sql.(*Rows).initContextClose in goroutine 1
	/Users/gopher/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.darwin-arm64/src/database/sql/sql.go:2996 +0x12c
```

As we can see, this code snippet causes panic in the `scanJSON` function and blocks the goroutine and the database connection. This case is difficult to catch because it does not return any error.

Related issue https://github.com/golang/go/issues/76465.

Yusufihsangorgel:
Scanning a JSONB column into a non-nil interface field routes through `scanJSONIntoInterface`, where `scanJSON` panicked on `dest.Addr()` because the interface element is not addressable, leaving the connection idle in transaction and the query blocked.

The check now sits in `scanJSON` and `scanJSONUseNumber`, the two scanners that call `dest.Addr()`, and in `scanNull`, which reached `dest.Set` on the same unaddressable element when the column was NULL. A dest that is already nil keeps scanning NULL as a no-op. Values that are unaddressable but still scannable, such as a pointer stored in an interface, keep dispatching to `PtrScanner`.

Fixes #1306

### DIFF
--- a/schema/scan_json_test.go
+++ b/schema/scan_json_test.go
@@ -1,0 +1,36 @@
+package schema
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Regression test for #1306: scanning JSON into a non-nil interface used to
+// panic on dest.Addr() because the interface element is not addressable. The
+// scan must fail with an error instead, so the query does not deadlock.
+func TestScanJSONUnaddressableInterface(t *testing.T) {
+	var value any = map[string]any{"a": float64(1)}
+	dest := reflect.ValueOf(&value).Elem()
+
+	fn := Scanner(dest.Type())
+	require.NotNil(t, fn)
+
+	require.NotPanics(t, func() {
+		err := fn(dest, []byte(`{"b":2}`))
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "nonaddressable")
+	})
+}
+
+func TestScanJSONAddressableMap(t *testing.T) {
+	var value map[string]any
+	dest := reflect.ValueOf(&value).Elem()
+
+	fn := Scanner(dest.Type())
+	require.NotNil(t, fn)
+
+	require.NoError(t, fn(dest, []byte(`{"b":2}`)))
+	require.Equal(t, map[string]any{"b": float64(2)}, value)
+}
--- a/schema/scan_json_test.go
+++ b/schema/scan_json_test.go
@@ -7,6 +7,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type jsonPayload struct {
+	A int `json:"a"`
+}
+
+type jsonRawScanner struct {
+	raw string
+}
+
+func (s *jsonRawScanner) Scan(src any) error {
+	b, err := toBytes(src)
+	if err != nil {
+		return err
+	}
+	s.raw = string(b)
+	return nil
+}
+
 // Regression test for #1306: scanning JSON into a non-nil interface used to
 // panic on dest.Addr() because the interface element is not addressable. The
 // scan must fail with an error instead, so the query does not deadlock.
@@ -14,14 +31,49 @@ func TestScanJSONUnaddressableInterface(t *testing.T) {
 	var value any = map[string]any{"a": float64(1)}
 	dest := reflect.ValueOf(&value).Elem()
 
-	fn := Scanner(dest.Type())
-	require.NotNil(t, fn)
-
+	var err error
 	require.NotPanics(t, func() {
-		err := fn(dest, []byte(`{"b":2}`))
-		require.Error(t, err)
-		require.Contains(t, err.Error(), "nonaddressable")
+		err = scanJSONIntoInterface(dest, []byte(`{"b":2}`))
 	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "nonaddressable")
+}
+
+// json_use_number fields scan through scanJSONUseNumber, which calls
+// dest.Addr() as well.
+func TestScanJSONUseNumberUnaddressable(t *testing.T) {
+	var value any = map[string]any{"a": float64(1)}
+	dest := reflect.ValueOf(&value).Elem().Elem()
+
+	var err error
+	require.NotPanics(t, func() {
+		err = scanJSONUseNumber(dest, []byte(`{"b":2}`))
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "nonaddressable")
+}
+
+// A pointer stored in an interface is not addressable either, but PtrScanner
+// scans into the pointer's element, which is. Such values scanned before the
+// fix and must keep scanning.
+func TestScanJSONPointerInInterface(t *testing.T) {
+	payload := &jsonPayload{}
+	var value any = payload
+	dest := reflect.ValueOf(&value).Elem()
+
+	require.NoError(t, scanJSONIntoInterface(dest, []byte(`{"a":7}`)))
+	require.Equal(t, 7, payload.A)
+}
+
+// A pointer type implementing sql.Scanner reaches its Scan method through
+// PtrScanner as well.
+func TestScanJSONScannerInInterface(t *testing.T) {
+	scanner := &jsonRawScanner{}
+	var value any = scanner
+	dest := reflect.ValueOf(&value).Elem()
+
+	require.NoError(t, scanJSONIntoInterface(dest, []byte(`{"a":7}`)))
+	require.Equal(t, `{"a":7}`, scanner.raw)
 }
 
 func TestScanJSONAddressableMap(t *testing.T) {
--- a/schema/scan_json_test.go
+++ b/schema/scan_json_test.go
@@ -39,6 +39,34 @@ func TestScanJSONUnaddressableInterface(t *testing.T) {
 	require.Contains(t, err.Error(), "nonaddressable")
 }
 
+func TestScanJSONIntoInterfaceNonAddressableNil(t *testing.T) {
+	var value any = map[string]any{"a": 1}
+	dest := reflect.ValueOf(&value).Elem()
+
+	require.NotPanics(t, func() {
+		err := scanJSONIntoInterface(dest, nil)
+		require.Error(t, err)
+	})
+}
+
+// Scanning NULL zeroes an addressable dest.
+func TestScanJSONAddressableNull(t *testing.T) {
+	value := map[string]any{"a": float64(1)}
+	dest := reflect.ValueOf(&value).Elem()
+
+	require.NoError(t, scanJSON(dest, nil))
+	require.Nil(t, value)
+}
+
+// A typed nil map inside an interface is already zero, so scanning NULL
+// into it stays a no-op.
+func TestScanJSONIntoInterfaceNilMapNull(t *testing.T) {
+	var value any = map[string]any(nil)
+	dest := reflect.ValueOf(&value).Elem()
+
+	require.NoError(t, scanJSONIntoInterface(dest, nil))
+}
+
 // json_use_number fields scan through scanJSONUseNumber, which calls
 // dest.Addr() as well.
 func TestScanJSONUseNumberUnaddressable(t *testing.T) {
--- a/schema/scan.go
+++ b/schema/scan.go
@@ -352,6 +352,9 @@ func scanJSON(dest reflect.Value, src any) error {
 	if src == nil {
 		return scanNull(dest)
 	}
+	if !dest.CanAddr() {
+		return fmt.Errorf("bun: Scan(nonaddressable %s)", dest.Type())
+	}
 
 	b, err := toBytes(src)
 	if err != nil {
@@ -364,6 +367,9 @@ func scanJSON(dest reflect.Value, src any) error {
 func scanJSONUseNumber(dest reflect.Value, src any) error {
 	if src == nil {
 		return scanNull(dest)
+	}
+	if !dest.CanAddr() {
+		return fmt.Errorf("bun: Scan(nonaddressable %s)", dest.Type())
 	}
 
 	b, err := toBytes(src)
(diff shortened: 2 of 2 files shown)


Aoang (commented):
This `CanAddr` check seems too restrictive when applied before scanner dispatch:

```go
dest = dest.Elem()
if !dest.CanAddr() {
    return fmt.Errorf("bun: Scan(nonaddressable %s)", dest.Type())
}

if fn := Scanner(dest.Type()); fn != nil {
    return fn(dest, src)
}
```

A non-addressable value is not necessarily unscannable. In particular, an interface may contain a non-nil pointer:

```go
type Payload struct {
    A int `json:"a"`
}

p := &Payload{}
var value any = p
```

After `dest = dest.Elem()`, `dest` represents the `*Payload` stored in the interface and may not be addressable. However, this worked previously because `Scanner(*Payload)` can dispatch to `PtrScanner`, which scans into the addressable `dest.Elem()`.

With the new check, we return `Scan(nonaddressable *Payload)` before `PtrScanner` gets a chance to handle it. The same issue may affect pointer types implementing `sql.Scanner`, since those scanners don't necessarily require `dest.CanAddr()` either.

I think the addressability check should stay in the scanners that actually call `dest.Addr()` (`scanJSON` / `scanJSONUseNumber`) rather than being enforced in `scanJSONIntoInterface` before scanner dispatch.

It would also be useful to add a regression test for a non-nil pointer stored in `any`, e.g. `var value any = &Payload{}`, to ensure this existing behavior remains supported.

Yusufihsangorgel:
Right on both counts. The `sql.Scanner` case is not just theoretical: on the old head, a Scanner stored in `any` hit the same nonaddressable error in my sqlite repro.

I moved the check out of the dispatch and into `scanJSON` and `scanJSONUseNumber`. Also added the regression test you suggested, `TestScanJSONPointerInInterface`, plus `TestScanJSONScannerInInterface` for the Scanner variant. Both fail on the revision you reviewed and pass now. The #1306 map case still errors instead of hanging the connection.

schema/scan_json_test.go
```diff
@@ -0,0 +1,88 @@
...
+	dest := reflect.ValueOf(&value).Elem()
+
+	var err error
+	require.NotPanics(t, func() {
+		err = scanJSONIntoInterface(dest, []byte(`{"b":2}`))
+	})
+	require.Error(t, err)
+	require.Contains(t, err.Error(), "nonaddressable")
+}
+
```
Aoang, line 41:
```suggestion

func TestScanJSONIntoInterfaceNonAddressableNil(t *testing.T) {
	var value any = map[string]any{"a": 1}
	dest := reflect.ValueOf(&value).Elem()

	require.NotPanics(t, func() {
		err := scanJSONIntoInterface(dest, nil)
		require.Error(t, err)
	})
}
```

A missed test case that needs to be fixed.

Aoang (changes requested):

schema/scan_json_test.go
Yusufihsangorgel: Applied your test unchanged, it fails without the fix and passes with it. On a nil src, `scanJSON` goes straight to `scanNull` before reaching the `CanAddr` guard, and the `dest.Set` there is what panics on the unaddressable map. I put the same check just before that `Set` and added `TestScanJSONAddressableNull` so NULL still zeroes addressable dests.

Aoang (approved):

PR #1396, opened 2026-07-22
fix(schema): prevent line-comment SQL injection with negative numbers (CVE-2024-44906 class)

### LINKED ISSUE #1395: CVE-2024-44906 fix was not applied to the query builder / schema formatter (still injectable on v1.2.18)
### Summary

CVE-2024-44906 (GHSA-h4h6-vccr-44h2) was fixed in `driver/pgdriver` in v1.2.15 by
padding negative numeric arguments so they can't form a `--` line comment when
placed immediately after a `-`. The **same flaw is still present in bun's core
query builder / `schema` formatter**, which is a separate code path the fix
didn't touch. It reproduces on the latest release (**v1.2.18**) and affects the
API most people use — `ColumnExpr`, `NewRaw`, `Where("… = ?", x)` — regardless of
the underlying driver (pgdriver, pgx, lib/pq).

I'm reporting this publicly because the mechanism is already documented (this CVE,
CVE-2024-27289 for pgx, and the Sonar write-up); this is an incomplete fix rather
than a novel issue.

### Root cause

Two independent formatters; only `pgdriver` got the guard:

| Path | Location | Guard? |
|------|----------|--------|
| Driver `$N` substitution | `driver/pgdriver/format.go` → `appendArg` | ✅ v1.2.15 |
| Query builder `?` substitution | `schema/querygen.go` → `QueryGen.Append` (int/float cases) | ❌ |
| Reflected model values | `schema/append_value.go` → `AppendIntValue` / `AppendFloatValue` | ❌ |

The builder emits a negative integer as a bare literal via `strconv.AppendInt`, so
`… - ?` with `-42` becomes `… --42`, i.e. a line comment.

### Reproduction (bun v1.2.18, any pg driver)

```go
db := bun.NewDB(sqlDB, pgdialect.New())
// tables: example(result int, name text)={7,'alice'}; secrets(result int, name text)={999,'TOPSECRET'}

payload := "foo\n 1 AND 1=0 UNION SELECT name FROM secrets; --"
var names []string
err := db.NewRaw("SELECT name FROM example WHERE result=-? OR name=?", -42, payload).Scan(ctx, &names)
// err == nil ; names == ["TOPSECRET"]   <- data from `secrets` exfiltrated
```

SQL received by the server (`log_statement=all`):

```
SELECT name FROM example WHERE result=--42 OR name='foo
 1 AND 1=0 UNION SELECT name FROM secrets; --'
```

`--42` comments out the intended predicate; the newline inside the (correctly
quoted) string argument resumes on a fresh line and runs the injected `UNION`.
The patched `pgdriver` path is not affected — it emits `result=- -42 …`.

### Preconditions

1. A negative numeric argument whose placeholder immediately follows `-`.
2. A user-controlled string argument later on the same line.
3. Both values attacker-influenced.

### Suggested fix

Port the `pgdriver.appendArg` guard into the `schema` formatter:

```go
func appendSignedInt(b []byte, v int64) []byte {
	if v < 0 && len(b) > 0 && b[len(b)-1] == '-' { // avoid accidental "--" comment
		b = append(b, ' ')
	}
	return strconv.AppendInt(b, v, 10)
}
```

Apply in `schema/querygen.go` `QueryGen.Append` (int/int32/int64 and float32/float64
cases) and `schema/append_value.go` (`AppendIntValue`, `AppendFloatValue`). A
driver-independent regression test:

```go
q := db.NewSelect().ColumnExpr("1000-?", -500)
b, _ := q.AppendQuery(db.QueryGen(), nil)
require.NotContains(t, string(b), "--") // e.g. "SELECT 1000- -500"
```

### Environment

- bun v1.2.18 (latest), PostgreSQL 16, reproduced with lib/pq, pgx, and pgdriver.

### References

- CVE-2024-44906 / GHSA-h4h6-vccr-44h2 (pgdriver fix)
- CVE-2024-27289 / GHSA-m7wr-2xf7-cm9p (pgx, same class)
- https://www.sonarsource.com/blog/double-dash-double-trouble-a-subtle-sql-injection-flaw/

namtzigla:
## What

Fixes a line-comment SQL injection in the query builder / `schema` formatter — the same class as **CVE-2024-44906**, whose fix landed in `driver/pgdriver` but was never applied to the `schema` formatter that powers `ColumnExpr`, `NewRaw`, `Where("… = ?", x)`, and reflected model values.

Closes #1395.

## Why

The `schema` formatter inlines numeric arguments as bare literals. A negative number formatted immediately after a `-` yields `--`, which PostgreSQL parses as a line comment. A later string argument containing a newline can then break out of the comment and inject SQL.

Reproduction on `master` / v1.2.18 (any driver — pgdriver, pgx, lib/pq):

```go
payload := "foo\n 1 AND 1=0 UNION SELECT name FROM secrets; --"
var names []string
db.NewRaw("SELECT name FROM example WHERE result=-? OR name=?", -42, payload).Scan(ctx, &names)
// names == ["TOPSECRET"]   <- data exfiltrated
```

SQL sent to the server:

```
SELECT name FROM example WHERE result=--42 OR name='foo
 1 AND 1=0 UNION SELECT name FROM secrets; --'
```

`--42` comments out the intended predicate; the newline in the (correctly quoted) string argument resumes on a new line and runs the injected `UNION`.

## Fix

Mirror the `driver/pgdriver.appendArg` guard: insert a space before a negative numeric literal when the preceding byte is `-`, so `1000-?` with `-500` formats as `1000- -500` instead of `1000--500`.

Applied in:
- `schema/querygen.go` — `QueryGen.Append` (`int`, `int32`, `int64`, `float32`, `float64`); `uint*` types are unsigned and left unchanged.
- `schema/append_value.go` — the reflected `AppendIntValue`, `AppendFloat32Value`, `AppendFloat64Value`.

A single helper `guardLineComment(b, negative)` is added next to `Append`.

## Tests

New driver-independent regression test `schema/querygen_linecomment_test.go` asserts that formatting `1000-?` with negative `int`/`int32`/`int64`/`float32`/`float64` never contains `--`. `go test ./schema/...`, `go vet ./schema/...`, and `gofmt` are clean.

## References

- CVE-2024-44906 / GHSA-h4h6-vccr-44h2 (pgdriver fix this ports)
- CVE-2024-27289 / GHSA-m7wr-2xf7-cm9p (pgx, same class)
- https://www.sonarsource.com/blog/double-dash-double-trouble-a-subtle-sql-injection-flaw/

### DIFF
--- a/schema/append_value.go
+++ b/schema/append_value.go
@@ -187,7 +187,7 @@ func AppendBoolValue(gen QueryGen, b []byte, v reflect.Value) []byte {
 }
 
 func AppendIntValue(gen QueryGen, b []byte, v reflect.Value) []byte {
-	return strconv.AppendInt(b, v.Int(), 10)
+	return strconv.AppendInt(guardLineComment(b, v.Int() < 0), v.Int(), 10)
 }
 
 func AppendUintValue(gen QueryGen, b []byte, v reflect.Value) []byte {
@@ -203,11 +203,11 @@ func appendUint64Value(gen QueryGen, b []byte, v reflect.Value) []byte {
 }
 
 func AppendFloat32Value(gen QueryGen, b []byte, v reflect.Value) []byte {
-	return dialect.AppendFloat32(b, float32(v.Float()))
+	return dialect.AppendFloat32(guardLineComment(b, v.Float() < 0), float32(v.Float()))
 }
 
 func AppendFloat64Value(gen QueryGen, b []byte, v reflect.Value) []byte {
-	return dialect.AppendFloat64(b, float64(v.Float()))
+	return dialect.AppendFloat64(guardLineComment(b, v.Float() < 0), float64(v.Float()))
 }
 
 func appendBytesValue(gen QueryGen, b []byte, v reflect.Value) []byte {
--- a/schema/querygen.go
+++ b/schema/querygen.go
@@ -43,6 +43,18 @@ func (f QueryGen) IdentQuote() byte {
 	return f.dialect.IdentQuote()
 }
 
+// guardLineComment inserts a space when a negative numeric literal is about to
+// be appended immediately after a '-'. Without it, formatting an expression
+// such as "col - ?" with a negative argument produces "col --1", where the "--"
+// starts a SQL line comment that can be abused for SQL injection. This mirrors
+// the fix applied to driver/pgdriver (CVE-2024-44906).
+func guardLineComment(b []byte, negative bool) []byte {
+	if negative && len(b) > 0 && b[len(b)-1] == '-' {
+		return append(b, ' ')
+	}
+	return b
+}
+
 func (gen QueryGen) Append(b []byte, v any) []byte {
 	switch v := v.(type) {
 	case nil:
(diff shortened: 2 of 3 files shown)


Aoang:
Thanks for the PR!

PR #1401, opened 2026-07-22
fix(pgdriver): guard protocol message sizes against overflow and invalid lengths

### LINKED ISSUE #1397: pgdriver: message-length uint32 overflow → SQL injection (pgx CVE-2024-27304 class) + unbounded server-controlled allocations
## pgdriver: message-length `uint32` overflow → SQL injection (same class as pgx CVE-2024-27304), + unbounded/negative server-controlled allocations (DoS)

`driver/pgdriver` writes Postgres protocol message/parameter length prefixes by casting `len(...)` to `uint32`/`int16` **with no size guard**, and reads server-supplied lengths into `make([]byte, n)` **with no cap or negative check**. The write side is the direct analogue of pgx **CVE-2024-27304** (integer overflow in message size → protocol desync → SQL injection); the read side is a memory-safety/DoS issue.

Reporting publicly because the write-side class is already public (pgx CVE-2024-27304) and this is the same unpatched pattern.

### 1. Write side — length prefix overflow → SQL injection (High, conditional)

`driver/pgdriver/write_buffer.go`:
```go
func (b *writeBuffer) FinishMessage() {
	binary.BigEndian.PutUint32(b.Bytes[b.msgStart:], uint32(len(b.Bytes)-b.msgStart)) // no guard
}
func (b *writeBuffer) FinishParam() {
	binary.BigEndian.PutUint32(b.Bytes[b.paramStart:], uint32(len(b.Bytes)-b.paramStart-4)) // no guard
}
```

pgdriver uses the simple protocol: `formatQuery` inlines all args into **one** `Query` message (`proto.go` `writeQuery`). When a single query reaches 4 GiB, `uint32(len)` wraps; the server reads a short length `W = realLen mod 2³²`, consumes `W-4` bytes as the statement, and parses the byte at offset `W` as the **start of a new protocol frame**. Those bytes are attacker-controlled (the tail of an oversized parameter), so the attacker can frame a new `Q` message with arbitrary SQL. Quote-escaping/`standard_conforming_strings` give no protection beyond the truncation boundary. `bytea` doubles via hex, so ~2 GB of user bytes suffices.

Preconditions match pgx's CVE: an app puts a single multi-GB user-controlled value into one statement (blob/file storage, bulk value list). Also `proto.go:834` `WriteInt16(int16(len(args)))` truncates >32767 bind params on the extended-protocol path (Bind desync), and `FinishParam` overflows for a single ≥4 GB bind parameter.

pgx fixed the equivalent by returning an error when the size exceeds `math.MaxUint32`; pgdriver has no such guard.

### 2. Read side — unbounded / negative allocations from server-supplied lengths (Medium, DoS)

Message length is read as a signed int32 and never validated (`proto.go` `readMessageType` → `int(l)-4`), then used directly:
- `proto.go:103` `make([]byte, n)` — huge `n` → multi-GB allocation (OOM).
- `proto.go:98` `r.buf[:n]` — negative `n` (server sends length < 4) → slice-bounds **panic**.
- `proto.go:480` `make([]byte, msgLen-4)` in the SASL path — reachable **pre-auth**; huge → OOM, `<4` → negative make panic.
- `column.go:64/110/134` `make([]byte, dataLen)` — column data length; `-1` (NULL) is handled, other negatives panic, huge → OOM.

Trigger: a malicious/compromised Postgres server, or an on-path attacker over the **unverified-TLS default** (see the separate TLS report), sending one crafted message with an oversized or sub-4 length field. No cap exists anywhere.

### Suggested fix (mirrors pgx)
- `FinishMessage`/`FinishParam`: error when `len-start > math.MaxUint32`.
- Guard `int16(len(args))` against `> math.MaxInt16` / `> 0xFFFF`.
- Read side: reject `msgLen`/`dataLen` that are negative (except the `-1` NULL sentinel) or exceed a sane cap before any `make([]byte, n)`.

### References
- pgx CVE-2024-27304 / GHSA-mrww-27vc-gghv (integer overflow in message size → SQLi)

### Environment
- bun/pgdriver upstream master (v1.2.18-era), reproduced by source analysis of the write/read paths. Happy to provide a large-payload PoC on request.

namtzigla:
## What

Guards the pgdriver Postgres protocol against message-size integer overflow (write
side) and invalid/negative server-supplied lengths (read side).

Closes #1397.

## Why

`driver/pgdriver` wrote message/parameter length prefixes by casting `len()` to
`uint32`/`int16` with no bound check. A single query or parameter ≥ 4 GiB wrapped the
length field, desyncing the server — the same integer-overflow → protocol-desync → SQL
injection class as jackc/pgx **CVE-2024-27304**. bun uses the simple protocol, so an
oversized inlined parameter goes out as one `Query` message and the wrapped length lets
the server parse attacker bytes as a new frame. Separately, server-supplied lengths were
fed straight into `make([]byte, n)` / slice expressions, so a malicious or on-path server
could crash the client (negative length → panic) or exhaust memory.

## Change

**Write side**
- `writeBuffer` records a sticky error when a message (`FinishMessage`) or parameter
  (`FinishParam`) would exceed the 32-bit size limit, and `Conn.write` refuses to send a
  buffer whose message failed to build — so an overflowing message never reaches the wire.
- `writeBindExecute` rejects `> math.MaxInt16` parameters instead of truncating the count.

**Read side**
- `readMessageType` rejects a length field `< 4` (would produce a negative body length).
- `reader.ReadTemp` rejects a negative `n`.
- `readColumnValue`'s default branch treats a non-positive length as NULL rather than
  `make([]byte, negative)`.

## Tests

`driver/pgdriver/message_size_test.go` covers the write guards (the 4 GiB limit is a
test-overridable var so the test needs no multi-GB allocation), the Bind param-count cap,
and the read-side rejections. `go build`, `go vet`, `gofmt` clean; new tests pass. (The two
pre-existing failing tests in this package require a live PostgreSQL and fail identically on
`master`.)

## Notes / scope

The read-side change guards the negative-length panics; a configurable cap for very large
(but positive) server-declared lengths is intentionally left out to avoid breaking legitimate
large result values — happy to add one if preferred.

## Reference
- jackc/pgx CVE-2024-27304 / GHSA-mrww-27vc-gghv

### DIFF
--- a/driver/pgdriver/column.go
+++ b/driver/pgdriver/column.go
@@ -61,6 +61,13 @@ func readColumnValue(rd *reader, dataType int32, dataLen int) (any, error) {
 		return readBytesCol(rd, dataLen)
 	}
 
+	if dataLen < 0 {
+		// dataLen == -1 signals SQL NULL; any other negative value is invalid
+		// and must not reach make([]byte, dataLen) (would panic). dataLen == 0
+		// is fine and keeps its previous behavior (empty value).
+		return nil, nil
+	}
+
 	b := make([]byte, dataLen)
 	if _, err := io.ReadFull(rd, b); err != nil {
 		return nil, err
--- a/driver/pgdriver/driver.go
+++ b/driver/pgdriver/driver.go
@@ -174,6 +174,14 @@ func (cn *Conn) reader(ctx context.Context, timeout time.Duration) *reader {
 }
 
 func (cn *Conn) write(ctx context.Context, wb *writeBuffer) error {
+	// Refuse to send a message that failed to build (e.g. it exceeded the
+	// protocol size limit); sending it would desync the connection.
+	if wb.err != nil {
+		err := wb.err
+		wb.Reset()
+		return err
+	}
+
 	cn.setWriteDeadline(ctx, -1)
 
 	n, err := cn.netConn.Write(wb.Bytes)
--- a/driver/pgdriver/error.go
+++ b/driver/pgdriver/error.go
@@ -84,4 +84,15 @@ func isBadConn(err error, allowTimeout bool) bool {
 
 var (
 	errRequiresParameter = errors.New("pgdriver: requires that standard_conforming_strings=on and client_encoding=UTF8")
+
+	// errMessageTooLarge is returned instead of sending a message whose length
+	// prefix would overflow the protocol's 32-bit size field. Without this guard
+	// the length wraps and the server desyncs, which can be abused for SQL
+	// injection (cf. jackc/pgx CVE-2024-27304).
+	errMessageTooLarge = errors.New("pgdriver: message size exceeds the maximum of 4 GiB")
+
+	// errInvalidMessageLength is returned when the server sends a message whose
+	// length field is smaller than the 4-byte length prefix itself (which would
+	// otherwise lead to a negative-length slice/allocation panic).
+	errInvalidMessageLength = errors.New("pgdriver: server sent a message with an invalid length")
 )
--- a/driver/pgdriver/message_size_test.go
+++ b/driver/pgdriver/message_size_test.go
@@ -1,0 +1,84 @@
+package pgdriver
+
+import (
+	"bytes"
+	"context"
+	"database/sql/driver"
+	"encoding/binary"
+	"errors"
+	"math"
+	"testing"
+)
+
+func TestFinishMessage_TooLargeSetsError(t *testing.T) {
+	orig := maxMessageSize
+	maxMessageSize = 8
+	defer func() { maxMessageSize = orig }()
+
+	wb := newWriteBuffer()
+	wb.StartMessage(queryMsg)
+	wb.WriteString("some query that exceeds the tiny test limit")
+	wb.FinishMessage()
+
+	if !errors.Is(wb.err, errMessageTooLarge) {
+		t.Fatalf("wb.err = %v, want errMessageTooLarge", wb.err)
+	}
+}
+
+func TestConnWrite_RefusesOversizedMessage(t *testing.T) {
+	wb := newWriteBuffer()
+	wb.setErr(errMessageTooLarge)
+
+	// wb.err is set, so write must return it without touching the (nil) netConn.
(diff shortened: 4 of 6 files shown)


driver/pgdriver/column.go
Aoang:
```suggestion
	if dataLen < 0 {
		if dataLen == -1 {
			return nil, nil
		}
		return nil, errInvalidMessageLength // TODO or nil?
	}
```

The negative-length validation should happen before the type switch. PostgreSQL defines `-1` as the only valid negative column length, representing SQL `NULL`; any value below `-1` is invalid protocol data.

In the current implementation, known column types return from the switch before reaching the new `dataLen < 0` check. As a result, a length such as `-2` may be silently converted into an empty string, zero, zero time, or empty byte slice, depending on the column type. Unknown types treat it as `NULL`. This also makes behavior inconsistent across types.

The later `dataLen < 0` check can then be removed. This preserves the valid `-1` NULL sentinel while consistently rejecting all other negative lengths.

Aoang (changes requested):

driver/pgdriver/column.go
```diff
@@ -31,7 +31,12 @@ const (
...
 func readColumnValue(rd *reader, dataType int32, dataLen int) (any, error) {
-	if dataLen == -1 {
+	if dataLen < 0 {
+		if dataLen == -1 {
+			return nil, nil
+		}
+		return nil, errInvalidMessageLength // TODO or nil?
+	}
 		return nil, nil
 	}
```
Aoang:
```suggestion
```

driver/pgdriver/column.go
```diff
@@ -61,6 +66,13 @@ func readColumnValue(rd *reader, dataType int32, dataLen int) (any, error) {
 		return readBytesCol(rd, dataLen)
 	}
 
+	if dataLen < 0 {
+		// dataLen == -1 signals SQL NULL; any other negative value is invalid
+		// and must not reach make([]byte, dataLen) (would panic). dataLen == 0
+		// is fine and keeps its previous behavior (empty value).
+		return nil, nil
+	}
+
```
Aoang:
```suggestion
```

Issue #1398, opened 2026-07-22
pgdriver: TLS certificate not verified by default; WithInsecure(false) still unverified; SCRAM channel binding disabled


namtzigla:
## pgdriver: TLS certificate not verified by default, `WithInsecure(false)` still doesn't verify, SCRAM channel binding disabled

`driver/pgdriver`'s default connection is **encrypted but the server certificate is never verified**, and there is **no option-based way to get a verified connection**. Combined with SCRAM channel binding being disabled, an on-path (active MITM) attacker can impersonate the database server, harvest credentials, and read/modify queries and results.

### 1. Unverified by default
`config.go` `newDefaultConfig()`:
```go
TLSConfig: &tls.Config{InsecureSkipVerify: true},
```
Every default connection performs the SSLRequest upgrade and then accepts **any** certificate (any/no chain, wrong host, self-signed, expired) — `ServerName` is also empty, so there is not even hostname checking. It does not silently fall back to plaintext (good — `enableSSL` errors if the server has no SSL), so it is *unauthenticated* TLS, not plaintext.

### 2. `WithInsecure(false)` is a footgun — it does NOT give a verified connection
`config.go`:
```go
func WithInsecure(on bool) Option {
	return func(conf *Config) {
		if on {
			conf.TLSConfig = nil                                  // plaintext
		} else {
			conf.TLSConfig = &tls.Config{InsecureSkipVerify: true} // TLS, but UNVERIFIED
		}
	}
}
```
A developer calling `WithInsecure(false)` to obtain a *secure* connection still gets a MITM-able one. The **only** ways to get verification are `WithTLSConfig(customVerifiedConfig)` or a DSN with `sslmode=verify-full`. There is no `With*` option that turns verification on.

### 3. SCRAM channel binding disabled
`proto.go` `authSASL`:
```go
case sasl.ScramSha256.Name:
	saslMech = sasl.ScramSha256
case sasl.ScramSha256Plus.Name:
	// ignore
```
[...]

PR #1402, opened 2026-07-22
fix(pgdriver): verify server certificate for WithInsecure(false)

namtzigla:
## What

Makes `pgdriver.WithInsecure(false)` produce a TLS connection that **verifies the server
certificate**, instead of one that silently skips verification.

Addresses the `WithInsecure(false)` part of #1398.

## Why

```go
func WithInsecure(on bool) Option {
	if on {
		conf.TLSConfig = nil                                  // plaintext
	} else {
		conf.TLSConfig = &tls.Config{InsecureSkipVerify: true} // TLS, but UNVERIFIED
	}
}
```

A developer calling `WithInsecure(false)` to get a *secure* connection still got one with
no server authentication — open to active man-in-the-middle. There was also no option-based
way to obtain a verified connection (only `WithTLSConfig` or DSN `sslmode=verify-full`).

## Change

- `WithInsecure(false)` now returns a verifying `*tls.Config` (`&tls.Config{}`,
  `InsecureSkipVerify: false`). `WithInsecure(true)` is unchanged (plaintext).
- `enableSSL` populates `ServerName` from the connection address when a verifying config
  doesn't set one (crypto/tls requires `ServerName` or `InsecureSkipVerify`).
- Doc comment + migration note: to intentionally skip verification (self-signed cert in
  dev), pass `WithTLSConfig(&tls.Config{InsecureSkipVerify: true})`.

## Compatibility

This is a **security-motivated behavior change**: code that relied on `WithInsecure(false)`
to connect to a server with an untrusted/self-signed certificate will now fail the TLS
handshake and must opt into skipping verification explicitly (see above). Flagging clearly
for review.

## Deliberately out of scope (separate follow-ups, want maintainer input)

- Changing the package **default** (`newDefaultConfig` still uses `InsecureSkipVerify: true`);
  flipping it would break local/dev connections to self-signed servers.
- Adding **SCRAM-SHA-256-PLUS** channel binding (currently ignored in `authSASL`); it needs
  wiring the TLS connection state into the SASL client and integration testing.
- Gating `AuthenticationCleartextPassword` behind verified TLS.

## Tests

`driver/pgdriver/config_tls_test.go`: asserts `WithInsecure(false)` yields a verifying
config and `WithInsecure(true)` yields plaintext, plus `hostWithoutPort` cases. `go build`,
`go vet`, `gofmt` clean.

### DIFF
--- a/driver/pgdriver/config.go
+++ b/driver/pgdriver/config.go
@@ -128,14 +128,34 @@ func WithTLSConfig(tlsConfig *tls.Config) Option {
 	}
 }
 
+// WithInsecure configures how the connection is protected.
+//
+//   - WithInsecure(true) disables TLS entirely (plaintext connection).
+//   - WithInsecure(false) enables TLS *with* server-certificate verification.
+//     ServerName is derived from the connection address at connect time.
+//
+// Note: previously WithInsecure(false) enabled TLS but skipped certificate
+// verification (InsecureSkipVerify: true), which left the connection open to
+// active man-in-the-middle attacks. To intentionally skip verification (e.g. a
+// self-signed certificate in development) pass an explicit config via
+// WithTLSConfig(&tls.Config{InsecureSkipVerify: true}).
 func WithInsecure(on bool) Option {
 	return func(conf *Config) {
 		if on {
 			conf.TLSConfig = nil
 		} else {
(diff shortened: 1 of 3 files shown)


Aoang (approved):
LGTM!

PR #1407, opened 2026-07-22
chore: replace go-hex with encoding/hex

### LINKED ISSUE #1385: Replace unmaintained github.com/tmthrgd/go-hex with stdlib encoding/hex
`github.com/tmthrgd/go-hex` has been unmaintained for ~7 years (last commit 2019). We'd love to use `bun` but our SBOM/supply-chain scanning flags it as an abandoned dependency. Since bun uses only two API-compatible functions from it afaik, I'd like to propose to replace it with the standard go library `encoding/hex`.

### Where it's used (quick search in the repo)  

The dependency is imported in `internal/hex.go` - aliased as `fasthex`, using:
- `fasthex.EncodedLen(...)`
- `fasthex.Encode(...)`
Every other `hex` reference in the codebase (again quick search in the repo) already uses stdlib `encoding/hex`. It's listed as a direct dependency in the root `go.mod` and as `// indirect` across the dialect/example modules.  

### Why we'd want to replace it  
 
- It's currently unmaintained: if a defect/vulnerability were found, there's no upstream to patch it. (There are no known CVEs today — currently this is _only_ a maintenance/supply-chain concern)
- A dependency could be droppen: rather than swapping one third-party lib for another - `encoding/hex` is maintained by the Go team and patched through normal Go releases.
- AFAIK it would be API- and behavior-compatible: `EncodedLen` and `Encode` have identical signatures and semantics, so the SQL output should be unchanged.  

### The only tradeoff I can see  

`go-hex` provides SIMD-accelerated encoding, so stdlib `encoding/hex` may be slower when serializing large `[]byte`/`bytea` blobs. I guess for most workloads this is is probably negligible, but maybe someone can include a benchmark on a representative blob size if maintainers want to quantify it before deciding.

Happy to open a PR if this direction is acceptable.

giri256:
## Summary

- replace the unmaintained `github.com/tmthrgd/go-hex` dependency with the standard library `encoding/hex`
- remove the dependency from the root module and every nested module
- add focused coverage for empty, single-write, and multi-write hex encoding

Closes #1385

## Validation

- `gofmt -d internal/hex.go internal/hex_test.go`
- `go test ./...`
- `go vet ./...`
- compile-only `go test -run '^$'` in every nested module
- verified the repository contains no remaining `tmthrgd/go-hex` references

The database-backed `driver/pgdriver` runtime tests were not run because no local PostgreSQL service was available; their compile-only check passed, and CI supplies the required database services.

### DIFF
--- a/internal/hex_test.go
+++ b/internal/hex_test.go
@@ -1,0 +1,43 @@
+package internal
+
+import "testing"
+
+func TestHexEncoder(t *testing.T) {
+	tests := []struct {
+		name   string
+		writes [][]byte
+		want   string
+	}{
+		{
+			name: "no writes",
+			want: "NULL",
+		},
+		{
+			name:   "single write",
+			writes: [][]byte{{0x00, 0x01, 0xfe, 0xff}},
+			want:   `'\x0001feff'`,
+		},
+		{
+			name:   "multiple writes",
+			writes: [][]byte{{0xab}, {0xcd}},
+			want:   `'\xabcd'`,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			enc := NewHexEncoder(nil)
+			for _, write := range test.writes {
+				if _, err := enc.Write(write); err != nil {
+					t.Fatalf("Write() error = %v", err)
+				}
+			}
+			if err := enc.Close(); err != nil {
+				t.Fatalf("Close() error = %v", err)
+			}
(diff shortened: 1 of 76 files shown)


internal/hex_test.go
Aoang, line 1: The standard `encoding/hex` library is very common, so I don't think we need to add unit tests for it. We can remove the `internal/hex_test.go` file.

Aoang (changes requested):

PR #1408, opened 2026-07-22
fix(internal): replace unmaintained go-hex with stdlib encoding/hex

### LINKED ISSUE #1385: Replace unmaintained github.com/tmthrgd/go-hex with stdlib encoding/hex

vipulbhasin23:
Fixes #1385

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/puzpuzpuz/xsync/v3 v3.5.1
 	github.com/rs/zerolog v1.34.0
 	github.com/stretchr/testify v1.8.1
-	github.com/tmthrgd/go-hex v0.0.0-20190904060850-447a3041c3bc
 	github.com/vmihailenco/msgpack/v5 v5.4.1
 )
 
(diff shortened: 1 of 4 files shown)


vipulbhasin23:
Hey @Aoang — noticed #1407 was opened shortly before this one for the same issue. Happy to defer to whichever you'd prefer to merge.

One difference worth flagging: this PR keeps the change scoped to the root module and `internal/hex.go`, since those are the only places with a real (non-`// indirect`) dependency on `go-hex` — the nested submodules just carry it as an indirect entry that should resolve naturally on their own next `go mod tidy`. #1407 additionally sweeps all the nested `go.mod`/`go.sum` files, which is also a valid approach if you'd rather have it done in one pass.

Let me know if you'd like any changes here, or if you'd prefer to go with #1407 instead.

vipulbhasin23:
Closing in favor of #1407, which also covers the nested go.mod/go.sum files.

PR #1412, opened 2026-07-24
fix(insert): keep field in VALUES when any slice element has a non-default value

### LINKED ISSUE #1394: Dropped field value(s) on slice model insert
in `query_insert.go` when you determine which fields you'll range over to insert values (and which you'll instead add to RETURNING clause), you filter out fields for insert based only on the first slice element value. see: https://github.com/uptrace/bun/blob/44c54f24205e2d10c6cff6d0541ad09f59afcf4c/query_insert.go#L436

so then any later models after the first element in the slice that actually have a non-zero value set for that field won't have it inserted. which is obviously quite a serious issue as it means data then doesn't get inserted in the database!

JoeJoeflyn:
Fixes #1394

`getFields()` inspected only the first slice element to decide whether a `NotNull` + `nullzero`/`default:` field should move to `RETURNING` instead of `VALUES`. If a later element had a real value, it was silently dropped and replaced by the column DEFAULT.

Now the field moves to `RETURNING` only when **every** element marshals to DEFAULT/NULL. Single-struct path unchanged.

```sql
-- Before: email dropped, second row's value lost
INSERT INTO t (name) VALUES ('zero-email'), ('with-email') RETURNING id, email
-- After:  email kept, both values persisted
INSERT INTO t (name, email) VALUES ('zero-email', 'none'), ('with-email', 'alice@example.com') RETURNING id, email
```

Tested on PostgreSQL (pg + pgx) and SQLite with `-race`. New regression test `TestIssue1394_DroppedFieldValueOnSliceInsert` included.


emoss08:
Nudge, dealing with this issue.

JoeJoeflyn:
?

> Nudge, dealing with this issue.

PR #1414, opened 2026-08-08
fix(bunotel): honor WithMeterProvider when reporting DB stats metrics

### LINKED ISSUE #1270: Meter provider option ignored by some metrics on bunotel
[github.com/uptrace/bun/extra/bunotel/otel.go:61](https://github.com/uptrace/bun/blob/26996878ae81ebe320e81511f9b35c0945f0df57/extra/bunotel/otel.go#L61) should be passing the meter provider (and trace provider) specified in the options collected at [github.com/uptrace/bun/extra/bunotel/otel.go:35](https://github.com/uptrace/bun/blob/26996878ae81ebe320e81511f9b35c0945f0df57/extra/bunotel/otel.go#L35).

The net result is during the Init call some metrics are forcibly recorded to the global meter provider instead of the configured one.

hsdfat:
QueryHook.Init called otelsql.ReportDBStatsMetrics without forwarding the configured meter provider, so the go.sql.connections_* instruments were always registered on the global meter provider even when the hook was built with WithMeterProvider.

Remember the provider on the hook and pass otelsql.WithMeterProvider when one was configured; hooks created without the option keep using the global provider as before.

Fixes #1270

### DIFF
--- a/extra/bunotel/option.go
+++ b/extra/bunotel/option.go
@@ -57,6 +57,7 @@ func WithTracerProvider(tp trace.TracerProvider) Option {
 func WithMeterProvider(mp metric.MeterProvider) Option {
 	return func(h *QueryHook) {
 		if mp != nil {
+			h.meterProvider = mp
 			h.meter = mp.Meter("github.com/uptrace/bun")
 		}
 	}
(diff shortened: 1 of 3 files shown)


Aoang (approved):
LGTM, Thanks for the contribution!

PR #1420, opened 2026-08-12
perf(schema): memoize LookupField for prefixed column names

eray-can:
## Problem

`Table.LookupField` is called once per column, per scanned row (`structTableModel.scanColumn`, `hasManyModel.Scan`). It has two paths:

```go
func (t *Table) LookupField(name string) *Field {
	if field, ok := t.FieldMap[name]; ok {
		return field                  // fast path: returns a shared *Field, no allocation
	}
	// slow path: walk StructMap, then...
	return field.WithIndex(index)     // clones the Field + allocates a new index slice
}
```

Any column name that addresses a joined/embedded struct — `author__name`, `league__country__translation__name` — misses `FieldMap` and takes the slow path. Those names are extremely common: every `Relation()` generates them, and applications add their own via `ColumnExpr("... AS x__y")`.

Because the result is recomputed for every row, a query that returns *N* rows with *M* prefixed columns performs *N×M* `Field` clones. Nothing about the result depends on the row — only on `(table, name)`.

## Change

Memoize the slow path in a `sync.Map` on `Table`.

This is safe: the fast path already hands out a shared `*Field` from `FieldMap`, and every `*Field` method is read-only with respect to the receiver (`ScanValue`, `AppendValue`, `Value`, … all operate on the passed `reflect.Value`). Sharing the resolved field is therefore no different from what `FieldMap` already does.

Negative results are cached too, so repeated lookups of an unknown column stay cheap.

## Benchmarks

Added `schema/table_lookup_bench_test.go`, using a `Book → Author → Country` belongs-to chain.

```
                                    ns/op      B/op     allocs/op
BenchmarkLookupFieldPrefixed-16      2411      2088          17     (before)
BenchmarkLookupFieldPrefixed-16       183         0           0     (after)

BenchmarkLookupFieldDirect-16          70         0           0     (before)
BenchmarkLookupFieldDirect-16          68         0           0     (after, unchanged)
```

~13× faster on the prefixed path, allocations gone. The direct path is untouched.

## Impact on a real application

Found while profiling a REST API backed by Bun + PostgreSQL. The endpoint lists 50 fixtures with several belongs-to and has-many relations (~620 rows scanned per request).

Heap profile over a 30-second load test:

| | total alloc | `Field.WithIndex` | GC cycles |
|---|---|---|---|
| before | 4.85 GB | 3.27 GB (65.9%) | 1077 |
| after | 1.28 GB | *not in profile* | 219 |

`Field.WithIndex` was the single largest allocation site in the whole application — larger than JSON encoding, the pgx driver and the HTTP layer combined. CPU time in `database/sql.convertAssignRows` dropped from 31.8% to 19.4%.

Endpoint responses were byte-for-byte identical across 18 scenarios (pagination, search, filters, two languages) before and after.

## Notes

- `go test ./schema/... .` and `./internal/...` pass.
- I could not run `-race` locally (no cgo toolchain on this machine); CI should cover it. `sync.Map` is used only for read-mostly caching, and a duplicate computation under a race would simply produce an equivalent `*Field`.
- The cache is bounded by the set of distinct column names a table is ever scanned with, which is bounded by the application's queries.

### DIFF
--- a/schema/table.go
+++ b/schema/table.go
@@ -422,8 +422,7 @@ func (t *Table) LookupField(name string) *Field {
 	}
 
 	if v, ok := t.lookupCache.Load(name); ok {
-		field, _ := v.(*Field)
-		return field
+		return v.(*Field)
 	}
 
 	field := t.lookupFieldSlow(name)
--- a/schema/table.go
+++ b/schema/table.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/jinzhu/inflection"
@@ -38,6 +39,11 @@ func SetTableNameInflector(fn func(string) string) {
 // Table represents a SQL table created from Go struct.
 type Table struct {
 	dialect Dialect
+
+	// lookupCache memoizes LookupField results for dotted/prefixed column
+	// names (e.g. "author__name"). Those names miss FieldMap and take the
+	// slow path below, which clones the Field on every call.
+	lookupCache sync.Map // map[string]*Field
 
 	Type      reflect.Type
 	ZeroValue reflect.Value // reflect.Struct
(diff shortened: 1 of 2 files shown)


schema/table.go
```diff
@@ -401,11 +407,31 @@ func (t *Table) addField(field *Field) {
...
+// already assumed for the FieldMap fast path above, so the value is safe to
+// share.
 func (t *Table) LookupField(name string) *Field {
 	if field, ok := t.FieldMap[name]; ok {
 		return field
 	}
 
+	if v, ok := t.lookupCache.Load(name); ok {
+		field, _ := v.(*Field)
+		return field
```
Aoang:
```suggestion
		return v.(*Field)
```

Aoang (commented):
LGTM

eray-can:
Applied, thanks. Only `*Field` values are ever stored in `lookupCache`, so the assertion can't fail — and a cached nil `*Field` still asserts cleanly, which keeps the negative-caching path working.

Re-ran the suite and the benchmarks after the change:

```
BenchmarkLookupFieldPrefixed-16    146.4 ns/op    0 B/op    0 allocs/op
BenchmarkLookupFieldDirect-16       54.5 ns/op    0 B/op    0 allocs/op
```

PR #1421, opened 2026-08-18
fix(schema): encode msgpack fields with the dialect binary literal (#1219)

ChrisJr404:
### Describe the pull request

Fields tagged with `bun:",msgpack"` did not round-trip on SQLite (and MySQL) — only on PostgreSQL.

`appendMsgpack` encoded the value straight through `internal.HexEncoder`, which always emits PostgreSQL's `'\x...'` bytea literal. SQLite and MySQL don't recognize that syntax, so the payload was stored verbatim as text and reading the row back failed with a msgpack decode error:

```
sql: Scan error on column index 1, name "encoded": msgpack: unexpected code=5c decoding map length
```

(`0x5c` is the leading `\` of the `\x` prefix that got stored as data.)

The fix encodes the msgpack payload into a buffer and delegates to `Dialect.AppendBytes`, the same routine bun already uses for regular `[]byte` columns. Each dialect now emits its own binary literal — `'\x...'` for PostgreSQL, `X'...'` for SQLite and MySQL. The PostgreSQL output is byte-for-byte identical to before, so existing behavior is unchanged.

`internal.HexEncoder` was the only consumer of that code and is now unused, so it's removed.

Link to the issue: https://github.com/uptrace/bun/issues/1219

### Tests

- `internal/dbtest`: `TestMsgpackRoundTrip` — create/insert/select round-trip of a `,msgpack` field, run for every dialect via `testEachDB` (verified green on SQLite locally).
- `schema`: `TestAppendMsgpack_DialectSpecificLiteral` — asserts the appended literal matches the dialect's `AppendBytes` and that PostgreSQL still produces the `'\x...'` form.

### Checklist

- [x] I agree to follow the Code of Conduct by submitting this pull request.
- [x] I have read and acknowledge the Contributing guide.
- [x] I have added test cases to cover the new code.


ChrisJr404:
The round-trip test was still failing on pg, mysql, mariadb and mssql because the value literal was only half the problem. A msgpack struct was still mapped to a JSON/text column, so inserting the binary literal got rejected. Pushed a follow-up that forces msgpack fields to a Blob column (bytea, blob, varbinary), so the column type matches the literal. Verified the sqlite round-trip passes locally and confirmed the generated DDL is a binary column on all four dialects.

PR #1422, opened 2026-08-19
fix: run BeforeSelect hook in Count and Exists

### LINKED ISSUE #1289: Using ScanAndCount the BeforeSelect Hook is not called when Getting the Count using Pagination.
I am using BeforeSelect to add a filter for tenants and when use ScanAndCount the COUNT SQL does not include the where clause for the Tenant added in `BeforeSelect` when the results are Paginated. 

Using the source attached you can see the tenant is not applied when calling ScanAndCount. Uses zerolog because that is how we found it. 

[example.zip](https://github.com/user-attachments/files/23178935/example.zip)

ocomsoft:
## Summary

- `Count()` and `Exists()` skip the `BeforeSelect` hook, so any model that applies query predicates from that hook (e.g. multi-tenant row filtering) silently returns unfiltered results
- Added the same `beforeSelectHook` guard that `Scan()` already uses to both `Count()` and `Exists()`
- Added tests verifying the hook fires and that a `BeforeSelect`-based tenant filter actually restricts the result set

## Details

`Scan()` calls `beforeSelectHook(ctx)` before building SQL (line 946), but `Count()` and `Exists()` go straight to `AppendQuery` without it. This means any model implementing `BeforeSelectHook` to add WHERE clauses — the documented pattern for multi-tenant scoping — gets no filtering on count/exists queries.

The fix adds the same `if q.table != nil { q.beforeSelectHook(ctx) }` guard to `Count()` and `Exists()`, matching `Scan()`'s existing behaviour.

### Tests added

1. **Hook event tests** — verify `BeforeSelect` fires for `Count` and `Exists` (added to existing `TestModelHook`)
2. **`TestBeforeSelectHookOnCountAndExists`** — multi-tenant integration test: a model applies `WHERE tenant_id = ?` in `BeforeSelect`, seeds two tenants with different row counts, and asserts that `Count` and `Exists` return tenant-scoped results

Fixes #1289


Aoang (commented):
`ScanAndCount` can execute `BeforeSelectHook` twice on the same query when using the sequential transaction path.

`scanAndCountSeq` first calls `q.Scan()`, then calls `q.Count()` on the same `SelectQuery`. With this change, both calls invoke `beforeSelectHook`, so any query modifications made by the hook—such as adding `WHERE` clauses or `JOIN`s—may be applied twice. This can lead to duplicated conditions, conflicting joins, or incorrect results for non-idempotent hooks.

Please consider using independent query clones for the scan and count operations, or otherwise preventing hook mutations from accumulating on the same query object. It would also be useful to add a test covering paginated `ScanAndCount` with `Conn(tx)`.

PR #1428, opened 2026-09-01
feat(schema): support Go 1.27 uuid.UUID

### LINKED ISSUE #1427: Go 1.27 uuid.UUID not supported: "bun: Scan(unsupported uuid.UUID)"
The `uuid.UUID` type introduced with go 1.27 appears to be unsupported.

Tried with both the latest release, and latest commit on master.

Go's `database/sql` appears to support it: https://go-review.googlesource.com/c/go/+/819361

> // Source values of type string and byte may be scanned into
> // values of type [*uuid.UUID]. The source may contain the string
> // representation of a UUID, or a 16-byte []byte.

https://cs.opensource.google/go/go/+/master:src/database/sql/sql.go;l=3377

It would be good to have support in bun as well

visheshgubrani:
Fixes #1427

Adds support for the Go 1.27 standard library `uuid.UUID` type.

## Changes

- Support `encoding.TextUnmarshaler` and `encoding.BinaryUnmarshaler` in schema scanning
- Support `encoding.TextAppender` and `encoding.TextMarshaler` when appending values
- Scan UUIDs from textual values, raw 16-byte slices, and `[16]byte` values
- Reject strings that are not valid textual UUIDs
- Support standard, compact, braced, and URN UUID formats
- Support UUID values and nil UUID pointers in PostgreSQL arrays
- Discover text-encoding types as `VARCHAR`
- Handle nonaddressable byte arrays when appending and checking zero values
- Add tests using the real Go 1.27 `uuid.UUID` type

The UUID scanning behavior matches `database/sql`. Raw UUID data is accepted from 16-byte binary values, while strings must contain a valid textual UUID.


jalavosus:
Hello! Just curious when this PR will land in `main`. Zero rush, I know you're doing this for free and your work is beyond appreciated. I implemented a small wrapper around `uuid.UUID` to be able to keep using `bun`, so it's no skin off my back if it takes you a bit. 

Thanks a lot!

Aoang:
Hi, thank you for contribution, and please forgive my late reply.

I’m concerned this may be a breaking change for existing types beyond `uuid.UUID`.

For example, a struct stored in a `jsonb` column may also implement `encoding.TextUnmarshaler`. Previously Bun decoded the stored JSON into that struct; with this change, `Scan` calls `UnmarshalText` instead. That method may expect a different format and fail on existing rows. Likewise, types that implement `TextAppender` or `TextMarshaler` may change from binary to text output and have a different inferred SQL type.

visheshgubrani:
Hey @Aoang, thanks for the review and no worries about the delay at all. You were right and the concern was broader than just `uuid.UUID`. This was changing behavior for any type with `UnmarshalText`, `MarshalText` or `AppendText`, so I've reworked it.

The thing I changed is that it's no longer a structural check. There's a `go1.27` build-tagged helper that detects the stdlib `uuid.UUID` type exactly, with a no-op fallback for older toolchains, so the module's go 1.24 minimum doesn't move. Anything that isn't that exact type is untouched. The cases you raised should behave as before now, a struct in a `jsonb` column decodes and appends as JSON again, `driver.Valuer` and `QueryAppender` keep their precedence over text methods and byte slices and byte arrays are back to the old dispatch. All of that has regression tests that fail on the old commit and pass now.

While fixing the issues you stated, I also found and sorted out a few things on the uuid path itself, `uuid[]` now parses elements as text so a 16-character token isn't mistaken for raw bytes, NULL elements are handled instead of failing, and `bun.NullZero(uuid.Nil())` no longer panics.

PR #1433, opened 2026-09-17
feat(schema): support per-dialect column types in struct tags

xue-ding-e:
Column types are not portable across databases: `tinyint` does not exist on postgres (first sync fails with `type "tinyint" does not exist`), `jsonb` does not exist on mysql, and every dialect has a few types like these. Until now the options were a portable lowest common denominator, or overriding the DDL outside of bun.

A field can now qualify its SQL type with a dialect name, and the dialect-specific type is used when creating the table on that dialect:

```go
type Model struct {
    Level   int16  `bun:"type:tinyint,pgtype:smallint,mssqltype:tinyint"`
    Payload string `bun:"type:text,mysqltype:json"`
}
```

- `Level` becomes `smallint` on postgres, `tinyint` on mysql/mssql
- `Payload` becomes `json` on mysql, `text` everywhere else
- dialects without a matching tag fall back to the generic `type`, and fields without these tags are completely unaffected

The option names are `<dialect>type:` with dialect names matching `dialect.Name().String()` for the bundled dialects: `pgtype`, `mysqltype`, `sqlitetype`, `mssqltype`, `oracletype`.

The resolution happens at DDL generation time (`CreateTableQuery.appendSQLType`), where the dialect is already known — that is the same place the existing Oracle `VARCHAR2` special case lives, and it composes correctly with pgdialect's autoincrement-to-serial rewriting (an explicit `pgtype:smallserial` wins over it).

Two notes on scope: `internal/dbtest` and the runtime (scanning, appending, transformers) are untouched — only the DDL type choice changes. And `migrate/sqlschema` inspectors intentionally compare the generic `CreateTableSQLType` (the package does not know the dialect by design), so until that is revisited, schema-diff tooling sees the generic type rather than the dialect-specific one.

The same feature exists elsewhere as prior art: ent has `SchemaType(map[string]string{dialect.MySQL: ..., dialect.Postgres: ...})` and SQLAlchemy has `TypeEngine.with_variant`, so the semantics here (dialect tag wins → generic type → bun default) follow an established pattern.

Testing:

- `schema` unit tests: tag parsing into `Field.DialectSQLTypes`, `Field.DialectSQLType` lookup
- create-table query tests: the same model produces the dialect-specific type per dialect (pg, mysql, sqlite, mssql) and falls back to the generic type where no tag matches
- `make test` equivalents locally: go vet + go test (+race) + linux/386 build across all modules, golangci-lint clean, and the `internal/dbtest` suite against postgres/mysql8/mysql5/mariadb/mssql containers with the same failures (timezone-sensitive snapshots on this machine only) as a clean checkout of master


xue-ding-e:
The earlier commitlint failure was on me — the first push used a non-conventional commit type. Rebased (code unchanged, now `feat(schema): ...`) so CI should go green on the new head.

Aoang:
In the original tag syntax:

- A key-value pair uses `key:value`, and the value is optional: `bun:"id,pk,alias:u"`.
- Multiple keys are separated by commas: `bun:"id,rel:belongs-to,join:profile_id=id"`.
- A value may contain `=`, as in the `join` option: `bun:"rel:belongs-to,join:profile_id=id"`.

For your proposed `bun:"type:tinyint,pgtype:smallint,mssqltype:tinyint"`, I suggest using `bun:"type:tinyint;pg=smallint;mssql=tinyint"` instead. This follows the conventional delimiter hierarchy: `:` > `,` > `;` > `=`.

xue-ding-e:
Thanks, Aoang. Updated in fd3fa19 to use your suggested syntax:

```go
Level int16 `bun:"type:tinyint;pg=smallint;mssql=tinyint"`
```

The separate `pgtype`/`mysqltype` options are removed. The type parser preserves parentheses and quoted SQL values, so cases such as `decimal(10,2)` and `enum('a;b','x=y')` remain intact. Ordinary comma-separated options and `join:profile_id=id` keep their existing parsing.

Added regression cases for all five dialect names, generic/inferred fallback and delimiter handling. Root tests, race tests, vet, lint and the dialect-module checks pass locally. The PR description now reflects this syntax and the validation performed. AI assistance was used for the changes, tests and this response.

